PR #579, opened 2018-10-06
Questionnaire Layout & Content Fixes

emerencia:
- [x] Fixed seed ordering to not give errors because one time responses were loaded before the depedent protocols.
- [x] Removed extra margins from admin preview of questionnaire view that made the preview have extra margins that the actual questionnaires did not.
- [x] When trying to submit a questionnaire with a group radio buttons, all of which are unchecked, it will now scroll back to that group of radio buttons, and display a line line to indicate its invalidity. Autoscrolling back up was already the case for selects, checkboxes, text inputs, and so on, just not for radios. And when filling out a long questionnaire, it can be very frustrating when you click submit and the thing just doesn't work.
- [x] Fixed email and mobile phone text inputs on small screen sizes to make use of the full width of the screen instead of half of it.
- [x] The company/application logos are now also displayed on the "Klaar" pagina.
- [x] Add an extra margin below submit buttons so that clicking the submit button doesn't require two presses on an iPhone X.
- [x] Added Mandy's text changes to the Evaluatiequestionnaire from the Google doc.
- [x] Limit the maximum increase in font-size caused by flow-text. Also removed flow-text from the labels of controls because that did nothing.
- [x] Added support for automatic rescaling of images in toasts and fixed resaling of the netherlands map images in the evaluatieonderzoek questionnaire.

Before and after view of the flow-text maximum font-size increase limit restriction:

chrome_2018-10-06_15-08-16

Nieuwe klaar pagina:
chrome_2018-10-06_16-49-35


app/views/application/_logos.html.haml
```diff
@@ -0,0 +1,7 @@
+.row
+  .col.s12
+    - if Rails.application.config.settings.logo.company_logo.present?
+      = image_tag Rails.application.config.settings.logo.company_logo, style: 'width: 20%; display: inline; vertical-align: middle'
+      = image_tag logo_image, style: 'width: 15%; display: inline; vertical-align: middle; margin-left:63.5%'
+    - else
+      = image_tag logo_image, style: 'width: 20%; margin: auto; display: block;', class: 'right-align'
```
frbl: ik wou m er de vorige keer al bij zetten, maar misschien nog even een `if logo_image.present?` hierbij. Voor als we geen logo boven aan de pagina willen

app/views/reward/index.html.haml
```diff
@@ -1,2 +1,3 @@
+= render partial: 'logos'
```
frbl: ow nice, ik wist niet eens dat een reward page dat niet had

spec/support/ui_macros.rb
```diff
@@ -15,9 +15,11 @@ def materialize_select(prompt, option, superelement = nil)
     # 2. Go to the parent ul
     # 3. Go to the parent div
     # 4. Click on it to open the dropdown
+    sleep(1)
```
frbl: :see_no_evil:

emerencia: Yep

app/views/reward/index.html.haml
emerencia: Ja het zag er allemaal wat kaal uit.

app/views/application/_logos.html.haml
emerencia: Ik vind m hip

emerencia: Fixed.

PR #978, opened 2019-07-16
update serializers gem

emerencia:


emerencia:
Fix with https://github.com/rails-api/active_model_serializers/compare/v0.10.9...v0.10.10#diff-5624f9cf35be4dc254412d979fd293c4R101

emerencia:
def json_key

PR #1107, opened 2019-11-06
Fb fixes to sdv branch

frbl:
# Description of feature
...

# Checklist


app/helpers/application_helper.rb
```diff
@@ -25,6 +26,17 @@ def logo_image
...
   private
 
+  def current_user_from_header
+    # current_auth_user is not defined if the including class does not
+    # include Knock::Authenticable. Hence we need to check if it is
+    # actually defined.
+    # @note We need to call this function first, and then fail. We cannot just
+    #   check if it is defined, as it is generated automatically.
+    current_auth_user&.person
+  rescue StandardError => _e
```
emerencia: What about all the errors that the JWT authentication procedure could throw? Are they some subclass of `StandardError`? Because then we are dropping a lot of good information here. Is there a way to check whether Knock::Authenticable was included? e.g., check what methods it defines and see if they're defined in the current context? Because I would prefer that over silencing all standard errors.

app/controllers/api/v1/jwt_api/auth_user_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+        end
+
+        private
+
+        def set_person
+          return if current_auth_user&.person.present?
+
+          result = { result: current_auth_user.to_json }
+          render(status: :not_found, json: result)
+        end
```
emerencia: Compare `app/controllers/api/v1/jwt_api/protocol_subscriptions_controller.rb` to this file. Here, you explicitly check if a current_auth_user has a person, and then render the method. There, you just return current_user (which in this case is the person of the current_auth user).

I think that the JwtApiController should implement a before filter that authenticates the user (i.e., it calls auth_user or authenticate_auth_user, or whatever is needed to trigger the authentication method), and then also checks that that the current_auth_user has a person (either with current_user or current_auth_user.person).

So then we no longer need to check whether a current_auth_user.person exists in any deriving subclass like we do here, and can no longer forget to check (like we do in `app/controllers/api/v1/jwt_api/protocol_subscriptions_controller.rb`).

app/controllers/api/v1/jwt_api/response_controller.rb
```diff
@@ -0,0 +1,92 @@
...
+          # token = request.headers['Authorization'].split.last
+          # Rails.logger.warn Knock::AuthToken.new(token: token)
+          # Rails.logger.warn current_auth_user.person.inspect
+
+          @person = current_auth_user&.person
+          return if @person.present?
+
+          result = { result: current_auth_user.to_json }
+          render(status: :not_found, json: result)
+        end
```
emerencia: This whole method can also go, see my earlier comment.

app/controllers/api/v1/jwt_api/response_controller.rb
```diff
@@ -0,0 +1,92 @@
...
+
+        def set_completed_responses
+          @responses = @person.my_completed_responses
+          return if @responses.present?
+
+          result = { result: 'Geen completed responses voor deze persoon gevonden' }
+          render(status: :not_found, json: result)
+        end
```
emerencia: I actually take issue with the fact that if there are no completed responses, that we throw a 404, instead of just returning an empty array with a 200. What are your thoughts on this? Same with the 'open responses' method below.

Because it's not like the user supplied an ID and we couldn't find an object with that ID. It's just that the user asked hey give me all my open responses, and we went sure, here you go: [] (empty array). I don't think it should give a 404 because it's just regular interaction.

I'm running into this problem with the ikia backend API that suddenly gives a 404 when I ask if there are questionnaires to be filled out and there are no new questionnaires to be filled out, which I find weird.

app/controllers/api/v1/jwt_api/response_controller.rb
```diff
@@ -0,0 +1,92 @@
...
+
+        def set_responses
+          @responses = @person.my_open_responses
+          return if @responses.present?
+
+          result = { result: 'Geen responses voor deze persoon gevonden' }
+          render(status: :not_found, json: result)
```
emerencia: This one ^

app/controllers/api/v1/jwt_api/protocol_subscriptions_controller.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    module JwtApi
+      class ProtocolSubscriptionsController < JwtApiController
+        def mine
+          render json: current_user.protocol_subscriptions, each_serializer: Api::ProtocolSubscriptionSerializer
```
emerencia: Shouldn't this be `current_auth_user.person`? Because if we use current_user, doesn't that open up the possibilities of accepting the other methods of authentication?

Or better still: ideally, we move the definition for current_user to the respective JwtApiController, JwtApiAndCookieController, BasicAuthController, and then let those definitions of current_user only exist of the allowed authentication methods in those respective classes.

It's not really a security concern, because to get into e.g., the basic auth controller you need valid basic auth, but if you also supply e.g., an Authentication header with a Bearer token, AND we happen to use current_user somewhere in a method of a basic auth controller (or deriving class), then a call to a basic_auth controller method will create an auth user and its person, just because we happened to pass a header along, which is probably something we don't want.

app/models/auth_user.rb
```diff
@@ -41,10 +41,21 @@ def team_from_payload(payload)
...
+      # Roles is also checked here for backwards compatibility
+      role_admin = metadata_from_payload(payload)['roles']&.include?(ADMIN_ACCESS_LEVEL)
+      access_level_admin = metadata_from_payload(payload)['access_level']&.include?(ADMIN_ACCESS_LEVEL)
+
+      if role_admin
+        ActiveSupport::Deprecation
+          .warn('roles should not be used in the payload anymore. '\
+                'This was renamed to access_level. '\
+                'Please update the payload.')
+      end
```
emerencia: nice

app/controllers/api/v1/jwt_api/response_controller.rb
frbl: good point, ive changed all of them to a 200

app/controllers/api/v1/jwt_api/protocol_subscriptions_controller.rb
frbl: hmm yeah I think this really depends on what we want. I like the fact that we have different ways of authenticating the user (that was originally how I created the api). And in that case, it is nice if there is just one function that we can use. What I did now is use current_user everywhere in the subclasses. We can split it up if we want later.

emerencia: Ok

emerencia: It's just that in most other classes I saw `current_auth_user.person` rather than `current_user`.

PR #1129, opened 2019-11-11
Add link_to_expandable functionality, fix layout of expandable sections, optimize db queries, fix ikia questionnaires

emerencia:
# Description of feature
...

# Checklist


lib/docker-entrypoint-webpack.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+# https://stackoverflow.com/a/38732187/1935918
+
+cd /app
+bundle install
+yarn install
```
emerencia: This is so we don't need to rebuild the whole image on a gem update or yarn.lock update.

projects/Differentiatie Binnenstebuiten/config/settings.yml
```diff
@@ -8,7 +8,7 @@ default: &default
   feature_toggles:
     allow_identifier_export: true
   logo:
-    favicon: 'projects/Differentiatie Binnenstebuiten/favicon.ico'
+    favicon: 'myfavicon.ico'
```
emerencia: I don't think this ever worked?

PR #1139, opened 2019-11-11
Fixed ando comments

frbl:
# Description of feature
...

# Checklist


app/controllers/api/v1/cookie_and_jwt_api/person_controller.rb
```diff
@@ -10,6 +10,8 @@ def me
           render json: current_user, serializer: Api::PersonSerializer
         end
 
+        delegate :my_students, to: :current_user
```
emerencia: What does this do exactly? Does this render current_user.my_students as a json? Without using the serializer for it? Is that what we want?

emerencia:
Verder lgtm

PR #1159, opened 2019-11-18
Refactor to use settings object

emerencia:
Instead of

```ruby
Rails.application.config.settings.logo.mentor_logo
```

now use:
```ruby
SETTINGS.logo.mentor_logo
```


emerencia:
Figure out why specs don't work

emerencia:
Not really needed and too finnicky to get to work.

PR #1160, opened 2019-11-19
Store results on base platform

emerencia:


app/models/response.rb
```diff
@@ -150,4 +154,12 @@ def response_expired?
     measurement.open_duration.present? &&
       Time.zone.now > TimeTools.increase_by_duration(open_from, measurement.open_duration)
   end
+
+  def generate_token
+    request_env = {}
+    Warden::JWTAuth::Hooks.new.send(:add_token_to_env,
```
frbl: Should this be in this class? Seems a bit weird to me. Maybe auth_user or person?

emerencia: Ok.

emerencia: Fixed.

PR #1162, opened 2019-11-21
Add method to export all responses

emerencia:
- [ ] Instead of changing the index, just add a new method and add it to the vsv api js as well.

### DIFF
--- a/app/controllers/api/v1/jwt_api/response_controller.rb
+++ b/app/controllers/api/v1/jwt_api/response_controller.rb
@@ -50,7 +50,7 @@ module Api
         end
 
         def set_responses
-          @responses = current_user.my_responses
+          @responses = current_user.my_open_responses
           return if @responses.present?
 
           render(status: :ok, json: [])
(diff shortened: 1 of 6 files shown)


app/controllers/api/v1/jwt_api/response_controller.rb
```diff
@@ -37,6 +42,13 @@ def set_completed_responses
           render(status: :ok, json: [])
         end
 
+        def set_all_responses
```
frbl: Set my responses?

config/webpacker.yml
```diff
@@ -54,6 +54,7 @@ development:
     headers:
       'Access-Control-Allow-Origin': '*'
     watch_options:
+      poll: 1000
```
frbl: What does this do?

app/controllers/api/v1/jwt_api/response_controller.rb
emerencia: Fixed. I also renamed the other methods for index and completed to be named consistently.

PR #1164, opened 2019-11-23
Create person from IKIA & fixes for the questionnaire integration with base-platform

emerencia:
# Description of feature
...

# Checklist


.env
```diff
@@ -46,3 +46,7 @@ IP_HASH_SALT='zout'
 INFO_SITE_URL='https://u-can-act.nl'
 
 PUSH_SUBSCRIPTION_URL='http://web:3000/api/v1/data/create_raw'
+BASE_PLATFORM_URL='http://localhost:3000'
+
+SHARED_SECRET='shared secret for generating email invite tokens'
+REGISTRATION_URL='http://localhost:3001/email-registratie'
```
frbl: this is the url for the questionnaire platform? (:3001?)

app/controllers/api/v1/jwt_api/people_controller.rb
```diff
@@ -0,0 +1,76 @@
...
+        before_action :set_email
+        before_action :check_email
+        before_action :set_parent
+        before_action :set_team_name
+        before_action :set_first_name
+        before_action :set_role_title
+
+        def create
+          @person = CreateChildPerson.run!(
+            parent: @parent,
```
frbl: nice, can we also use this for mentors / students? I like the notion of parent child, as its way more generic (even for this specific ikea (:-)) case

app/logic/jwt_authenticator.rb
```diff
@@ -19,7 +19,7 @@ def auth(cookies, params)
 
     def token_from_cookie_or_params(params, cookies)
       if token_from_params(params)
-        JWT.decode(token_from_params(params), Knock.token_public_key, true)
+        JWT.decode(token_from_params(params), Knock.token_public_key, true, algorithms: %w[RS256 HS256])
```
frbl: the algorithm is in the secret somewhere, right?

app/models/auth_user.rb
```diff
@@ -14,6 +14,14 @@ def jwt_subject
     auth0_id_string
   end
 
+  def generate_token
+    request_env = {}
+    Warden::JWTAuth::Hooks.new.send(:add_token_to_env,
```
frbl: can't we put this on a general place? WardenHelper or w/e, so we can use that one?

app/use_cases/create_anonymous_user.rb
```diff
@@ -29,38 +30,58 @@ def create_or_find_auth_user(auth0_id_string, access_level)
...
+        return auth_user
+      end
+    end
+
+    create_new_person(auth_user)
+  end
+
+  def create_new_person(auth_user)
+    ActiveRecord::Base.transaction do
+      auth_user.person = Person.create!(first_name: auth_user.auth0_id_string,
```
frbl: might be saying something stupid here, but won't `auth_user.person.create!` work? in that case you also don't need the transaction (I cannot test it here on my windows machine..)

app/use_cases/create_anonymous_user.rb
```diff
@@ -29,38 +30,58 @@ def create_or_find_auth_user(auth0_id_string, access_level)
...
+          auth_user.save!
+        end
+        return auth_user
+      end
+    end
+
+    create_new_person(auth_user)
+  end
+
+  def create_new_person(auth_user)
```
frbl: create a function `upsert_person`  in `auth_user`?

.env
emerencia: Yep. this is the quest-front-end stuff.

app/controllers/api/v1/jwt_api/people_controller.rb
emerencia: Yes, we can. It just depends on the team names and roles you specify.

app/logic/jwt_authenticator.rb
emerencia: Yeah, you're right. I had some decoding error and then I changed this line and restarted the server and it worked again, but I think it was the restarting that made it work, not the change of this line.

I'll revert this change.

app/use_cases/create_anonymous_user.rb
emerencia: No that doesn't work because `auth_user.person` is `nil`. `auth_user.create_person` should work, however. I'll change it to that.

chrome_2019-12-03_16-04-15

app/use_cases/create_anonymous_user.rb
emerencia: What does upsert person mean?

app/models/auth_user.rb
emerencia: Yes, but also no. I think it's too tied to this class.

You would have to move the jwt_subject function as well, but that is too closely tied to the model.  We're generating a token for an auth_user here, and we're not generating a token (we can't generate a token) for any other model in the database. And we're not duplicating this code anywhere. So I think this belongs in auth_user.

PR #1181, opened 2019-12-03
List children

emerencia:
- Add parent as explicit attribute to person


db/schema.rb
```diff
@@ -114,7 +114,9 @@
     t.string "iban"
     t.string "ip_hash"
     t.boolean "account_active", default: false, null: false
+    t.bigint "parent_id"
     t.index ["mobile_phone"], name: "index_people_on_mobile_phone", unique: true
+    t.index ["parent_id"], name: "index_people_on_parent_id"
```
emerencia: TODO: the version number in the schema didn't update because I made the migration after the current migration. So rename the migration so that it is after the current one, so the migration actually runs on currently running systems.

emerencia: Fixed.

app/controllers/api/v1/jwt_api/people_controller.rb
```diff
@@ -23,8 +25,47 @@ def create
...
+          if @child.valid?
+            render json: @child.result, serializer: Api::ChildSerializer, status: :ok
+          else
+            render json: @child.errors, status: :bad_request
+          end
+        end
+
+        def destroy_child
+          if @child.account_active?
+            # Don't actually destroy the child's account, just make us no longer the parent.
```
frbl: hmm this seems weird. Why do it like this? Wouldn't we want to make this more explicit? Perhaps because of gdpr reasons where we actually want to destroy someone?

app/models/person.rb
```diff
@@ -182,4 +185,10 @@ def warn_for_multiple_mentors
     Rails.logger.warn "[Attention] retrieving one of multiple mentors for student: #{id}" if
     ProtocolSubscription.active.where(filling_out_for_id: id).where.not(person_id: id).count > 1
   end
+
+  def not_own_parent
+    return if id.blank? || parent_id.blank? || id != parent_id
+
+    errors.add(:parent, 'cannot be parent of themselves')
```
frbl:
```suggestion
    errors.add(:parent, 'cannot be parent of yourself')
```

app/serializers/api/child_serializer.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module Api
+  class ChildSerializer < ActiveModel::Serializer
+    type 'people'
```
frbl: cool

app/serializers/api/child_serializer.rb
```diff
@@ -0,0 +1,12 @@
...
+    attributes :id, :first_name, :email, :account_active, :role
+
+    def role
```
frbl: are you sure this works (as in, that this takes preference over object.role? (I guess it does, but just to be sure))

app/use_cases/subscribe_to_protocol.rb
```diff
@@ -21,26 +21,32 @@ class SubscribeToProtocol < ActiveInteraction::Base
...
   # - start_date: the date when the subscription should start
+  # rubocop:disable Metrics/AbcSize
   def execute
+    the_mentor = mentor
     the_protocol = find_protocol
     the_start_date = find_start_date
     Rails.logger.warn("Protocol #{the_protocol.id} does not have any measurements") if the_protocol.measurements.blank?
 
-    create_or_find_protocol_subscription(the_protocol, the_start_date)
+    # If we are a child, our parent is our mentor
```
frbl: not a big fan of this tbh, but I see why you're doing it. Would it make sense to put this logic on the 'getter' for the mentor of a person? So we leave the field nil if there is no mentor, and we fill it with the mentor id only if we have a mentor? Not sure, but then we can still make a distinction (perhaps its completely unnecessary). What do you think?

app/use_cases/update_child_person.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  def find_team
+    team = person.role.team
+    return rerr('Team not found') if team.blank?
+
+    team
+  end
+
+  def rerr(msg)
```
frbl: didn't you have a general function for this?

app/controllers/api/v1/jwt_api/people_controller.rb
emerencia: I was thinking more along the lines of that we would never want to allow for a person to destroy someone else's account, no matter the circumstances. 

I can ask the IKIA people about what they think about this.

app/serializers/api/child_serializer.rb
emerencia: idd

app/serializers/api/child_serializer.rb
emerencia: Yes. I also test this in the specs.

app/use_cases/subscribe_to_protocol.rb
emerencia: I think if you think back to the u-can-act student protocols. They had one protocol subscription, and it was filling_out_for their mentor. The reason being that in the questionnaire, we could then evaluate the {{mentor_name}} (oid) and so on, to make the questionnaires more personal.

I think the same applies here. The only reason we're setting the parents as mentor is that so we can maybe in the future use their name to make things more personal. I was thinking of e.g. just storing the first name of the parent, so in the subject line of the invitation mail to the kids we can write something like, "Your parent/caregiver, {{mentor_name}}, has invited you to join IKIA!," which could lead to higher response rates.

I think what you're saying makes more sense however, it would just require a little more effort on my part, which is probably why I didn't think of it :p I.e., just add a {{parent_name}} variable and have it evaluate to `person.parent.name`.

app/use_cases/update_child_person.rb
emerencia: No that was the logg function.

app/use_cases/subscribe_to_protocol.rb
emerencia: Fixed.

I just removed setting the parent as mentor for now. We can add the {{parent}} variable later when we actually use it.

PR #1210, opened 2019-12-19
Fix security?

emerencia:
I don't think this works because it still has the older versions of serialize-javascript in the yarn.lock. So maybe remove those and then see if they get added when you do a yarn install, and if not, then it works.


emerencia:
This does not fix it.

PR #1283, opened 2020-01-14
Redis mutex locks for auth_user creation and protocol subscription AND distributions for questionnaires

emerencia:
#### TODO
- [x] Add a method to app/controllers/api/v1/jwt_api/questionnaire_controller to export the distribution 
of a questionnaire.
- [x] write a spec for the rake task itself
- [x] make realtime statistics optional
- [x] fix specs
- [x] move the || to the declaration of @ distribution in the questionnaire controller


README.md
```diff
@@ -353,6 +353,8 @@ The `content` attribute of a `Questionnaire` is a serialized array that stores t
 For all questions, it is allowed to use HTML tags in the texts. 
 Also, you may use any of the special variables defined in the previous section.
 
+All questions now support a `combines_with` attribute. The value of this attribute should be an array of (other) questionnaire IDs. This is used to indicate to the distributions engine that an additional histogram, combining the values of the question and the ones that it combines with, should also be calculated.
```
frbl: niet iets met 'conditional'? Aangezien dat precies is wat het is, een conditionele distributie.

app/controllers/api/v1/jwt_api/questionnaire_controller.rb
```diff
@@ -5,13 +5,19 @@ module V1
...
+        before_action :check_distributions_enabled, only: %i[distribution]
+        before_action :set_questionnaire, only: %i[show distribution]
+        before_action :set_distribution_and_check_min_responses, only: %i[distribution]
 
         def show
           # TODO: Add different formats
           render json: @questionnaire, serializer: Api::QuestionnaireSerializer
         end
 
+        def distribution
```
frbl: maybe move to a results controller?

frbl: perhaps as a subcontroller somehow; `questionnaire/1/results/distribution` or something

app/controllers/api/v1/jwt_api/questionnaire_controller.rb
```diff
@@ -32,13 +38,30 @@ def check_admin_authenticated
...
 
+        def set_distribution_and_check_min_responses
+          @distribution = RedisService.get("distribution_#{@questionnaire.key}") || '{}'
+
+          distribution = JSON.parse(@distribution)
+          return if distribution.present? &&
+                    distribution['total'].present? &&
+                    distribution['total'] >= Rails.application.config.settings.distribution_export_min_responses
+
+          render(status: :forbidden, json: 'Number of completed responses for this questionnaire is below the minimum')
```
frbl: forbidden sounds like a weird response here. wouldn't 404 be better? And why not just return the `{}`?

app/helpers/distribution_helper.rb
```diff
@@ -0,0 +1,73 @@
+# frozen_string_literal: true
+
+module DistributionHelper
+  VALUE = '_'
```
frbl: ? Perhaps add a comment? I'm guessing this is like an anonymous variable?

app/helpers/distribution_helper.rb
```diff
@@ -0,0 +1,73 @@
...
+    @questionnaire_content = questionnaire.content
+    range_questions(@questionnaire_content) + other_questions(@questionnaire_content)
+  end
+
+  def range_questions(questionnaire_content)
+    rg_instance = RangeGenerator.new
+    questionnaire_content
+      .select { |question| question[:type] == :range }
+      .map do |question|
+      rg_instance.send(:range_slider_minmax, question).merge(id: question[:id].to_s,
```
frbl: why private?

app/helpers/distribution_helper.rb
```diff
@@ -0,0 +1,73 @@
...
+                                                             type: question[:type],
+                                                             combines_with: question[:combines_with],
```
frbl: perhaps nest the 'combines_with' in a `:score` block.

app/helpers/distribution_helper.rb
```diff
@@ -0,0 +1,73 @@
...
+      { id: question[:id].to_s, type: question[:type], combines_with: question[:combines_with] }
+    end
+  end
+
+  def initialize_question(question, value, distribution)
+    qid = question[:id]
+    return if distribution[qid].present? && question[:type] == :range
+
+    distribution[qid] ||= {}
+    distribution[qid][value] ||= { VALUE => 0 } unless question[:type] == :range
```
frbl: merge the unless's?

app/helpers/distribution_helper.rb
```diff
@@ -0,0 +1,73 @@
...
+    return unless question[:type] == :range
+
+    pos = question[:min]
+    while pos <= question[:max]
+      distribution[qid][pos.to_s] = { VALUE => 0 }
+      pos += question[:step]
+    end
```
frbl:
```suggestion
    (question[:min]..question[:max]).step(question[:step]) { |pos| distribution[qid][pos.to_s] = { VALUE => 0 } }
```

frbl: (please test this)

app/jobs/calculate_distributions_job.rb
```diff
@@ -0,0 +1,28 @@
...
+    Questionnaire.find_each do |questionnaire|
+      CalculateDistribution.run!(questionnaire: questionnaire)
+    end
+  end
+
+  def clean_up_redis
+    RedisService.keys.each do |key|
+      next unless key.match?(/^distribution_/)
+
+      RedisService.del(key) unless Questionnaire.where(key: key[13..]).count.positive?
```
frbl: why 13? Prehaps add comment or constant

app/models/auth_user.rb
```diff
@@ -32,18 +32,21 @@ def from_token_payload(payload)
...
-      auth_user = CreateAnonymousUser.run!(
-        auth0_id_string: id,
-        team_name: team,
-        role_title: role,
-        access_level: access_level,
-        email: email
-      )
+      auth_user = nil
+
+      RedisMutex.with_lock("CreateAnonymousUser:#{id}") do
```
frbl: would it make sense to add this to the .run! itself? As in, we create some sort of decorator that actually mutexes a use_case (like we did for the redis caching).

frbl: hmm ah well, essentially that's what you're doing here, only a new level of abstraction

app/controllers/api/v1/jwt_api/questionnaire_controller.rb
emerencia: Good point. That would keep the front-end from breaking. If there are not enough responses, it would just show like an empty distribution instead of an error.

emerencia: Ok the vsv_api gem gets very confused if we return anything except 200, so I return an empty array with a total 0 and an error message now.

app/controllers/api/v1/jwt_api/questionnaire_controller.rb
emerencia: Fixed.

app/helpers/distribution_helper.rb
emerencia: So I wouldn't have to write specs for it. :p I will now.

emerencia: Fixed.

app/helpers/distribution_helper.rb
emerencia: This is just the default value for the structure. Imagine the structure not being a hash but just a single value, it would be this value. The reason that everything has to be a hash is so that we can nest combined scores into combined histograms that are constructed recursively. In order for that to work, here we choose a value that is short (because everything is stored in JSON format) and is probably not taken by any of the question IDs.

emerencia: Added the above comment to the code.

app/helpers/distribution_helper.rb
emerencia: I think the score block will be something different as we discussed. This is purely for combining histograms.

app/helpers/distribution_helper.rb
emerencia: Fixed. Thanks

app/helpers/distribution_helper.rb
emerencia: Ok i will write a spec for this method and then try your implementation and see if it still passes.

emerencia: Yep it seems to work identical. Fixed.

emerencia: Nice.

app/jobs/calculate_distributions_job.rb
emerencia: It's because it's the length of `distribution_`

emerencia: I added a comment.

app/models/auth_user.rb
emerencia: Yep. I kind of like that it's obvious. It's like the ActiveRecord::Base.Transaction blocks, but better.

PR #1309, opened 2020-02-03
Score calculation

emerencia:
Numericality mappings will be a separate pull request.


app/controllers/questionnaire_controller.rb
```diff
@@ -84,18 +84,39 @@ def check_interactive_content
     render(status: :bad_request, json: { error: 'Please supply a json string in the content field.' })
   end
 
+  # rubocop:disable Metrics/AbcSize
```
frbl: damn, this function actually looks quite daunting. you're sure it does not make sense to split it up?

app/models/questionnaire.rb
```diff
@@ -1,14 +1,29 @@
...
   validates :content, presence: true
   validates :key, presence: true, uniqueness: true, format: { with: /\A[a-z]+[a-z_0-9]*\Z/ }
-  serialize :content, Array
-  validate :all_content_ids_unique
-  validate :all_questions_have_types
-  validate :all_questions_have_titles
-  validate :all_questions_have_ids
+  serialize :content # Don't specify Hash type because otherwise databases with existing questionnaires won't work
+  validate :questionnaire_structure, if: -> { content.present? }
+  with_options if: :content_has_questions do
```
frbl: wow nice

app/models/questionnaire.rb
```diff
@@ -47,31 +89,92 @@ def all_content_ids_unique
...
+  def all_scores_have_nonempty_ids
+    result = content[:scores]
+             .select { |score| score[:ids].blank? }
+             .map { |score| score[:label] || score[:id] }
+    return if result.blank?
+
+    errors.add(:content, "the following scores have an empty ids attribute: #{result.pretty_inspect}")
+  end
+
+  def all_scores_use_existing_ids
```
frbl: is there a way to move this stuff out of this class? perhaps by using concerns or something? not sure

app/use_cases/enrich_content.rb
```diff
@@ -0,0 +1,80 @@
+# frozen_string_literal: true
+
+# Enrich response content (= questionnaire answers) with scores
+class EnrichContent < ActiveInteraction::Base
+  EnrichMissingDataError = Class.new(StandardError)
```
frbl: move to separate class?

app/use_cases/enrich_content.rb
```diff
@@ -0,0 +1,80 @@
...
+
+  hash :content, strip: false # strip: false means allow all keys
+  hash :questionnaire, strip: false
+
+  # Enriches the questionnaire outcomes with scores
+  #
+  # @param content [Hash] the response values as a hash (keys are strings)
+  # @param questionnaire [Hash] the questionnaire definition hash (keys can be symbols)
+  def execute
+    @questionnaire = questionnaire # work with instance variables instead of activeinteraction methods
```
frbl: why?

app/controllers/questionnaire_controller.rb
emerencia: I'll fix it.

app/models/questionnaire.rb
emerencia: Yep, it saves a lot of duplicated code in checks.

app/use_cases/enrich_content.rb
emerencia: Out of laziness because otherwise my IDE freaks out. Also there was some weird edge case if the value was nil, .blank? would return false or something.

I'll put it back and see if it causes the specs to fail.

app/use_cases/enrich_content.rb
emerencia: Ok.

app/use_cases/enrich_content.rb
emerencia: Seems to still be okay, so, fixed.

app/models/questionnaire.rb
emerencia: This says that is an example of concerns used wrong: https://medium.com/@carlescliment/about-rails-concerns-a6b2f1776d7d

Because it just moves the code and makes it less explicit.

emerencia: Because the functionality is tied to this class, it doesn't make sense to use Composition or Aggregation either.

I'll move them to a helper.

emerencia: Actually, this makes them non-private methods by default, unless we jump through a lot of hoops and declare them all private individually. I'll just leave it as is for now.

app/controllers/questionnaire_controller.rb
emerencia: Fixed.

PR #1323, opened 2020-02-20
Add preprocessing for scores

emerencia:


app/models/questionnaire.rb
```diff
@@ -184,4 +187,28 @@ def all_scores_have_known_operations
...
+    result = content[:scores]
+             .select { |score| score.key?(:preprocessing) }
+             .select { |score| invalid_preprocessing_steps?(score[:preprocessing]) }
+             .map { |score| score[:label] || score[:id] }
+    return if result.blank?
+
+    errors.add(:content, "the following scores have invalid preprocessing steps: #{result.pretty_inspect}")
+  end
+
+  def invalid_preprocessing_steps?(score_preprocessing)
```
frbl: would it not make sense to change this to something like `only_valid_preprocessing_steps?`?

app/use_cases/calculate_scores.rb
```diff
@@ -131,4 +135,13 @@ def round_result(value, score)
 
     value.round(score[:round_to_decimals])
   end
+
+  def preprocess_value(numeric_value, qid, score)
+    return numeric_value unless score.key?(:preprocessing) && score[:preprocessing].key?(qid)
+
+    preprocessing_step = score[:preprocessing][qid]
+    numeric_value *= str_or_num_to_num(preprocessing_step[:multiply_with]) if preprocessing_step.key?(:multiply_with)
```
frbl: it would be even nicer if you could create like a hash of function / key or something in the known preprocessing steps constant (something like `{name: :multiply_with, function: '*='}` or however you would do that in ruby. In that case you only have to keep one list, while now you have to keep this one consistent with the other list, right?

emerencia: I see what you mean. How can you get a reference to a method in ruby though?

emerencia: Found it.

emerencia: Ok this actually cleanes up the code a whole lot.

emerencia: Thanks for this. The code is awesome now.

emerencia: Fixed.

app/models/questionnaire.rb
emerencia: And then change it to .reject instead of .select, right? Sure, done.

PR #1342, opened 2020-02-26
Fix for the OTR redirect

frbl:
@emerencia would this be enough? I did not really have a way to test it properly, apart from the specs.


emerencia:
Lgtm so far, but to add:
- The token authentication(?) controller should redirect to questionnaires controller show instead of index, and pass the uuid of a random response from the invitationset of it's params[:q]
- The next page finder should be changed so that if the response someone just filled out is an OTR, it shouldn't do anything except redirect to a callback if one was provided. In particular, it should never redirect to another questionnaire.

emerencia:
@frbl 10 specs gaan nog fout

PR #1380, opened 2020-03-03
Only send invites when not filled out the daily questionnaire that day

emerencia:


projects/sport-data-valley/seeds/protocols/daily_protocol.rb
```diff
@@ -36,7 +36,7 @@
 
   general_daily_measurement.open_from_offset = offset.days + start_time
   general_daily_measurement.period = 1.week # every day for each questionnaire
-  general_daily_measurement.open_duration = 2.days # Open for two days
+  general_daily_measurement.open_duration = 1.days # Open for one day
```
frbl: ehm I'm not sure if this is correct. The idea is that a person can fill out this questionnaire also the next day. Or did you discuss it with Stephan?

emerencia: He reported an error that when someone filled out their training log, they would still receive a notification for the training log to be filled out that day. This is because right now the training logs are overlapping: the period is 1 day, but the open duration is 2 days, so there's always two open and we have no clue which one will be filled out when you click a link. I think to remove the confusion just have the previous one expire when the new one starts. Then there's  still always a daily measurement for people to fill out. 

Besides, they can also fill it out the next day, as long as it is before 1pm. Maybe that is what he meant?

PR #1441, opened 2020-03-27
Diary study support for IKIA

emerencia:


app/controllers/api/v1/cookie_and_jwt_api/protocol_subscriptions_controller.rb
```diff
@@ -11,6 +11,20 @@ def show
...
+        # If a start_time is specified (and no start_date is specified),
+        # the start_date is chosen as the given start_time added to midnight the next day.
+        def create
+          result = SubscribeToProtocol.run!(
+            protocol_name: protocol_subscription_create_params[:protocol_name],
+            person: current_user,
+            start_date: start_date,
+            only_if_not_subscribed: true
+          )
+          SendInvitations.run
```
frbl: is this necessary? Can't we just wait for the job to run? (this is fine too, but I like the idea that you don't have to do this manually)

app/controllers/api/v1/cookie_and_jwt_api/protocol_subscriptions_controller.rb
```diff
@@ -11,6 +11,20 @@ def show
           render json: @protocol_subscription, serializer: Api::ProtocolSubscriptionSerializer
         end
 
+        # Specify either a start_date or a start_time. A given start_date takes precedence.
+        # If a start_time is specified (and no start_date is specified),
+        # the start_date is chosen as the given start_time added to midnight the next day.
+        def create
+          result = SubscribeToProtocol.run!(
```
frbl: it would be very nice to also do this async tbh, especially since you don't have to return anything, just the 204. We could add a subclass to activeinteraction with a `run_async` functin or soething, that automatically runs the run! in a background job.

app/controllers/api/v1/cookie_and_jwt_api/protocol_subscriptions_controller.rb
```diff
@@ -25,6 +39,29 @@ def set_protocol_subscription
...
+        def start_time
+          return Time.zone.now if no_start_time_given
+
+          next_midnight = TimeTools.increase_by_duration(Time.zone.now.beginning_of_day, 1.day)
+          TimeTools.increase_by_duration(next_midnight, protocol_subscription_create_params[:start_time].to_i)
+        end
+
+        def no_start_time_given
+          protocol_subscription_create_params[:start_time].blank? ||
+            (protocol_subscription_create_params[:start_time].to_i.zero? &&
```
frbl: why is this? This seems a bit weird, you do it .to_i and check if it is 0, but it cannot be '0'? Perhaps add a comment why you do it like this.

app/controllers/api/v1/cookie_and_jwt_api/protocol_subscriptions_controller.rb
```diff
@@ -25,6 +39,29 @@ def set_protocol_subscription
...
+              protocol_subscription_create_params[:start_time] != '0')
+        end
+
+        def protocol_subscription_create_params
+          params.permit(:protocol_name, :start_date, :start_time)
+        end
```
frbl: on my device the indentation seems weird on the next lines

spec/controllers/api/v1/cookie_and_jwt_api/protocol_subscriptions_controller_spec.rb
```diff
@@ -61,4 +61,59 @@
       end
     end
   end
+
+  describe 'create' do
+    describe 'without cookie' do
+      it 'returns 401 if no protocol_subscription is available' do
```
frbl: why should this be a 401 btw? Isnt that something like unauthorized? Wouldnt 400 be better?

frbl: if no cookie is available, right?

app/controllers/api/v1/cookie_and_jwt_api/protocol_subscriptions_controller.rb
```diff
@@ -11,6 +11,20 @@ def show
           render json: @protocol_subscription, serializer: Api::ProtocolSubscriptionSerializer
         end
 
+        # Specify either a start_date or a start_time. A given start_date takes precedence.
+        # If a start_time is specified (and no start_date is specified),
+        # the start_date is chosen as the given start_time added to midnight the next day.
+        def create
```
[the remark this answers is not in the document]
emerencia: Yep, but that was in the basic auth API, and this was here. There is some overlap, but that one is different because it requires a mentor, and every protocol subscription to be started with that mentor as mentor. The two methods share code (in particular, they call the same use case), so I thought it was ok like this.

app/controllers/api/v1/cookie_and_jwt_api/protocol_subscriptions_controller.rb
emerencia: Alright, I'll try it.

emerencia: Ok I actually do a request for the status of the protocol subscription right after making one, so it's important that this happens synchronously. I'll leave this for now

app/controllers/api/v1/cookie_and_jwt_api/protocol_subscriptions_controller.rb
emerencia: Fixed.

app/controllers/api/v1/cookie_and_jwt_api/protocol_subscriptions_controller.rb
emerencia: It's because .to_i on a random string (or empty string) simply returns 0 instead of raising an error. I'll add a comment.

ConEmu64_2020-03-30_18-36-15

emerencia: Fixed.

app/controllers/api/v1/cookie_and_jwt_api/protocol_subscriptions_controller.rb
emerencia: I reformatted it, but it didn't change anything. Weird.

spec/controllers/api/v1/cookie_and_jwt_api/protocol_subscriptions_controller_spec.rb
emerencia: Yes, sorry the spec was named wrong.

emerencia: Yes

PR #1505, opened 2020-04-07
Forgery robustness

emerencia:
1. als csrf_failed -> roep niet de push subscriptions aan
1a. if csrf_failed -> don't calculate statistics for this response
1b. if csrf_failed -> we do calculate scores for this response
2. als values een lege hash is, ignore dan het request om die response aan te maken
3. ignore het request om een response te completen als de opened_at null is

The above controller checks exclude API calls to create questionnaire responses. They are only in effect for the regular questionnaires_controller (where were render a page showing the questionnaire for users).

### DIFF
--- a/app/controllers/questionnaire_controller.rb
+++ b/app/controllers/questionnaire_controller.rb
@@ -16,6 +16,7 @@ class QuestionnaireController < ApplicationController
   before_action :check_informed_consent, only: [:show]
   before_action :set_questionnaire_content, only: [:show]
   before_action :set_create_response, only: %i[create create_informed_consent]
+  before_action :check_opened_at, only: [:create]
   before_action :check_content_empty, only: [:create]
   before_action :check_content_hash, only: [:create]
   before_action :check_interactive_content, only: %i[interactive_render]
@@ -217,6 +218,13 @@ class QuestionnaireController < ApplicationController
     return if quest_content.present? && (quest_content.keys - [Response::CSRF_FAILED]).present?
 
     render(status: :bad_request, html: 'Cannot store blank questionnaire responses', layout: 'application')
+  end
+
+  def check_opened_at
+    return if @response.opened_at.present?
+
+    render(status: :bad_request, html: 'Cannot accept answers for an unopened response',
+           layout: 'application')
   end
 
   def check_informed_consent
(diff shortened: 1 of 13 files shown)


app/controllers/questionnaire_controller.rb
```diff
@@ -211,6 +213,20 @@ def check_content_hash
     end
   end
 
+  def check_content_empty
+    quest_content = questionnaire_content
```
frbl, line 217: actually quite misleading that we have 'questionnaire_content' and 'set_questionnaire_content' which do quite different things.

app/controllers/questionnaire_controller.rb
```diff
@@ -211,6 +213,20 @@ def check_content_hash
...
+    return if quest_content.present? && (quest_content.keys - [Response::CSRF_FAILED]).present?
+
+    render(status: :bad_request, html: 'Cannot store blank questionnaire responses', layout: 'application')
+  end
+
+  def check_opened_at
+    return if @response.opened_at.present?
+
+    render(status: :bad_request, html: 'Cannot accept answers for an unopened response',
```
frbl, line 226: hehe nice.

app/use_cases/recalculate_scores.rb
```diff
@@ -24,7 +24,9 @@ def execute
 
   def recalculate_response_ids(response_ids)
     ResponseContent.where(:id.in => response_ids).pluck(:content, :id).each do |content, rcid|
-      next if content.nil? # Can theoretically happen
+      # Note that we don't check for csrf_failed here (on purpose)
+      # It doesn't hurt to calculate scores for csrf_failed responses.
```
frbl, line 28: are you sure? it would hurt the calculation in terms of invalid data right?

emerencia: The score calculation has it built-in that if certain scores are missing, it won't calculate the total score, if you set `require_all` to `true`. If you don't set `require_all`, it will still calculate the scores as long as there is at least one of the scores used present.

What I meant was is that it doesn't affect the distributions since the scores are calculated just for the response itself, and affect only the response itself.

app/controllers/questionnaire_controller.rb
emerencia: I still have no clue how they did this in the first place. The only way I see possibled is if they used the API to read out the list of questionnaires.

emerencia: Either way, this fixes it.

app/controllers/questionnaire_controller.rb
emerencia: I think it's because this controller does both response stuff and questionnaire stuff. It should really be two separate controllers. This is on my list of things to fix, but refactoring is a nightmare and there's other stuff that takes priority right now.

PR #1566, opened 2020-05-12
Add support for questionnaires in multiple languages and translate th…

emerencia:
…e training log questionnaire

# Description of feature
...

# Checklist


app/generators/questionnaire_generator.rb
```diff
@@ -8,10 +8,17 @@ def initialize
...
   # rubocop:disable Metrics/ParameterLists
-  def generate_questionnaire(response_id:, content:, title:, submit_text:, action:, unsubscribe_url:, params: {})
+  # rubocop:disable Metrics/AbcSize
+  def generate_questionnaire(response_id:, content:, title:, submit_text:,
+                             action:, unsubscribe_url:, locale:, params: {})
     params[:response_id] = response_id
+    params['content[locale]'] = locale
     response = Response.find_by(id: response_id) # allow nil response id for preview
-    raw_content = content[:questions].deep_dup
+    raw_content = QuestionnaireTranslator.translate_content(content[:questions].deep_dup, 'i18n')
```
frbl, line 17: so essentially we don't translate it if it has an i18n attribute, right? We just use that key.

app/generators/questionnaire_generator.rb
```diff
@@ -8,10 +8,17 @@ def initialize
...
+    # Translate a second time to get rid of any remaining translation hashes because the i18n translation is optional
+    raw_content = QuestionnaireTranslator.translate_content(raw_content,
+                                                            Rails.application.config.i18n.default_locale.to_s)
+    content = QuestionnaireTranslator.translate_content(content, locale)
```
frbl, line 21: is it necessary to go through the list again? And why use the default locale on the line before this one, and not the passed in locale?

app/logic/questionnaire_translator.rb
```diff
@@ -0,0 +1,35 @@
...
+
+        return translate_content_aux(content[@locale.to_sym]) if content.key?(@locale.to_sym)
+
+        content.keys.each do |key|
+          content[key] = translate_content_aux(content[key])
+        end
+        return content
+      end
+      if content.is_a?(Array)
+        content.each_with_index do |value, idx|
```
frbl, line 27: can't you do a `return content.map{|x| translate_content_aux(x)}` here? It might not work well with the recursion, not sure.

app/generators/questionnaire_generator.rb
emerencia: We do translate it, we translate it to the i18n language.

app/generators/questionnaire_generator.rb
emerencia: Yes it is necessary because the languages that you're translating to might be different.

It is important that the raw content is always in the same language, because it defines the keys in the response hash, e.g. v1_brood for a checkbox question with an answer named pizza. We wouldn't want to see a v1_bread column appear in the CSV export when that represents the same answer, just in a different language. The v1 keys for checkboxes are determined by their raw labels, hence we make sure that the raw labels are language independent (note: this is not the label that is shown to the user, it is merely used for the name/id of the elements on the page).

We translate the raw content twice because the i18n language is optional, and if a translated string only has an 'nl' and 'en' component, it will be left untranslated in the string, which will give errors down the line when it expects a string and finds a hash. This is why after using the i18n translated strings wherever we can, we do a second pass to translate any remaining untranslated strings to the default language.

The content needs its own parsing step, since there we don't want any i18n labels.

emerencia: I added a comment to clarify this.

app/logic/questionnaire_translator.rb
emerencia: Yes, I think that's equivalent.

emerencia: I liked explicitly assigning things to content, but since we're returning that value as well, we don't need to modify it in place.

emerencia: Fixed.

PR #1606, opened 2020-06-09
UCF project configuration + schedule questionnaires on a specific day + prioritize questionnaire responses

emerencia:
- [x] u-can-feel project (including teams, protocols, and questionnaires)
- [x] have measurements fall on a certain day
- [x] have questionnaires scheduled at the same time be shown in a specified order


app/models/measurement.rb
```diff
@@ -83,7 +97,24 @@ def open_till(end_date)
...
+      new_start_date = start_date.beginning_of_day
+      while WEEKDAYS[new_start_date.wday] != open_from_day || open_from_with_offset(new_start_date) < start_date
+        # We go to the next day, but if it's daylight savings time switches, the next day can be more or less
+        # than 24 hours away, so to be sure we first go to noon the next day, and then back to the beginning
+        # of that day.
+        new_start_date = TimeTools.increase_by_duration(new_start_date, 36.hours).beginning_of_day
+      end
+      return open_from_with_offset(new_start_date)
+    end
+    open_from_with_offset(start_date)
```
frbl, line 114:
```suggestion
    return open_from_with_offset(start_date) if open_from_day.blank?
    new_start_date = start_date.beginning_of_day
    while WEEKDAYS[new_start_date.wday] != open_from_day || open_from_with_offset(new_start_date) < start_date
      # We go to the next day, but if it's daylight savings time switches, the next day can be more or less
      # than 24 hours away, so to be sure we first go to noon the next day, and then back to the beginning
      # of that day.
      new_start_date = TimeTools.increase_by_duration(new_start_date, 36.hours).beginning_of_day
    end
    open_from_with_offset(new_start_date) 
```

app/models/person.rb
```diff
@@ -104,28 +104,38 @@ def my_delegated_protocol_subscriptions
...
   end
 
+  # For any method that only returns open responses, we want them to be sorted by descending priority first,
+  # and ascending open_from second. This is because we call .first on this method to determine which is the next
+  # response that should be shown to the user.
   def my_open_responses(for_myself = true)
     active_subscriptions = protocol_subscriptions.active if for_myself.blank?
     active_subscriptions ||= my_protocols(for_myself)
-    active_subscriptions.map { |prot| prot.responses.opened_and_not_expired }.flatten.sort_by(&:open_from)
+    active_subscriptions.map { |prot| prot.responses.opened_and_not_expired }.flatten.sort_by(&:priority_sorting_metric)
```
frbl, line 113: don't you need to add sorting for the open from within the priority queue too? so, somehow, sort by priority, and then (taking that priority into account) sort by open_from?

frbl: right, nvm, you did that in the next part.

projects/u-can-feel/config/settings.yml
```diff
@@ -0,0 +1,22 @@
+# Beware! This file is public by definition
+default: &default
+  default_team_name: 'UCF'
+  metadata_field: 'https://u-can-feel.nl'
+  # This settings is included only for the specs. Don't remove
+  test_setting: 'test123'
+  project_title: 'U-can-feel'
```
frbl, line 7: u-can-feel instead of U-can-feel?

app/models/person.rb
emerencia: Yep. The priority sorting metric just means priority first, open from second.

PR #1643, opened 2020-07-07
Add fix for questionnaire names

frbl:
# Description of feature
...

# Checklist


projects/sport-data-valley/seeds/questionnaires/daily_questionnaire.rb
```diff
@@ -146,6 +146,6 @@
   questionnaire.key = "#{File.basename(__FILE__)[0...-3]}_#{day}"
 
   questionnaire.content = { questions: current_dagboek_content, scores: [] }
-  questionnaire.title = ''
+  questionnaire.title = "#{day.capitalize} vragenlijst"
```
emerencia, line 149: - leave this title empty
- add translation_key property to questionnaires
- make sure translation_key property is exposed by questionnaire and/or response serializer
- adjust base platform questionnaire roaster so when it sets the title for an event it checks for the translation_key and if it is not nil it uses that, and otherwise the title, and otherwise the questionnaire.key.humanize (these last two conditions is how it works currently, the first one with translation key is new)
- add the translation key to the i18n files of the base platform
- adjust the frontend of the questionnaire manager view so that it also prefers showing translated translation_keys over titles when listing protocol subscriptions

emerencia:
Btw, you can ignore my comment and should probably revert your change about the fix for empty questionnaire names: this has been fixed in a different branch already (i.e., it will use the humanized form of the questionnaire key if no title exists), and the questionnaire title is now sent back to the base platform when it requests an update of running protocol subscriptions, so please don't change that code now. 

The other changes to the squash questionnaire and stuff LGTM

projects/sport-data-valley/seeds/questionnaires/daily_questionnaire.rb
emerencia, line 149:
```suggestion
  questionnaire.title = ''
```

emerencia:
LGTM

emerencia (approved):
lgtm

PR #1645, opened 2020-07-08
Questionnaire fixes

emerencia:
In this pull request:

- Protocol subscriptions can now have an external identifier (nullable string)
- Allow protocol subscriptions to be listed via an external identifier (which will be a group id in the case of SDV)
- Allow protocol subscriptions to be created with a specific external identifier
- Allow protocol subscriptions to be canceled via the BasicAuth API
- Next page finder (for non mentors) now also redirects to responses filled out for a different user
- Next page finder no longer redirects to OTR responses
- My open responses in serialized person now lists all open responses except OTRs.
- Protocol subscription serializer now exposes the `first_name` attribute (which is set to the profile ID in the case of SDV)
- Protocol subscription serializer now lists the humanized key for questionnaires that have no title
- Response serializer now lists the external identifier (in the case of SDV, this is a group ID, so this can now be used by the u-can-act roaster/brewer to identify for which group the response was filled out).
- The daily protocol for Sport Data Valley now always starts the correct questionnaires on the correct days of the week, including the Sunday questionnaire to always start on a Sunday.


app/controllers/api/v1/jwt_api/protocol_subscriptions_controller.rb
```diff
@@ -8,11 +8,6 @@ def mine
           render json: current_user.my_protocols(true),
                  each_serializer: Api::ProtocolSubscriptionSerializer
         end
-
-        def delegated_protocol_subscriptions
```
frbl, line 12: we don't want this? Like a list of all protocolsubscriptions you have running for others? (I'm thinking Mentordashboard in react kindof stuff)

app/models/person.rb
```diff
@@ -94,23 +94,22 @@ def my_protocols(for_myself = true)
     prot_subs = protocol_subscriptions.active.joins(
       :protocol
     ).joins(
-      'FULL JOIN one_time_responses ON one_time_responses.protocol_id = protocols.id'
+      'LEFT JOIN one_time_responses ON one_time_responses.protocol_id = protocols.id'
```
frbl, line 97: added a spec for this to prevent it in the future?

projects/sport-data-valley/seeds/protocols/daily_protocol.rb
```diff
@@ -50,9 +62,10 @@
 
 sunday_measurement = protocol.measurements.find_by(questionnaire_id: questionnaire_id)
 sunday_measurement ||= protocol.measurements.build(questionnaire_id: questionnaire_id)
-sunday_measurement.open_from_offset = 6.days + start_time # open on Sunday
+sunday_measurement.open_from_offset = start_time # open on Sunday
+sunday_measurement.open_from_day = 'sunday'
 sunday_measurement.period = 1.week # every sunday
-sunday_measurement.open_duration = 2.days # Open for two days
+sunday_measurement.open_duration = 1.days # Open for two days
```
frbl, line 68:
```suggestion
sunday_measurement.open_duration = 1.days # Open for one day
```

spec/controllers/api/v1/basic_auth_api/protocol_subscriptions_controller_spec.rb
```diff
@@ -81,4 +86,67 @@
...
+    end
+  end
+
+  describe '#destroy' do
+    before do
+      basic_auth 'admin', 'admin'
+    end
+    let!(:protocol_subscription) { FactoryBot.create(:protocol_subscription, external_identifier: external_identifier) }
+
+    it 'cancel a protocol subscription' do
```
emerencia, line 127:
```suggestion
    it 'cancels a protocol subscription' do
```

app/controllers/api/v1/jwt_api/protocol_subscriptions_controller.rb
emerencia: Well the thing is that this was only used for the SDV platform, and now the SDV platform doesn't use it anymore. So when we do make a route that will be used, I'm in favor of adding it back, but I'm not in favor of adding methods right now just because we think they might be used later because what often happens is when that time arrives we'll have forgotten all about this method and just write a new one and then we have another chunk of unused code.

app/models/person.rb
emerencia: Because of the where conditions that we put in later down the line, it returns the same exact result and functionally, these queries are identical. It's just that I don't want to have to look at this later and wonder why we use a FULL join instead of a LEFT one, when just using a LEFT one should have been enough.

PR #1692, opened 2020-08-22
Add value to range_generator and docs

rikhuijzer:
# Description of feature
Add default value `value` to the range generator. This can be used to set the initial position of the range. 

The code can be made more pretty by passing the keywords as a collection. Such a workaround is necessary since we want to avoid setting the `value` keyword if `question[:value]` is `nil`, so that the range is automatically set to the center.

# Checklist
- [X] (If applicable) added example values of new ENV variables to .env.local and explained them in README.md.


app/generators/range_generator.rb
```diff
@@ -33,6 +33,18 @@ def range_slider(question)
                      max: minmax[:max].to_s,
                      step: question[:step] || 1,
                      required: true)
+    # TODO: Make this code prettier.
+    unless question[:value].nil?
```
emerencia, line 37:
```suggestion
    if question[:value].present?
```

app/generators/range_generator.rb
```diff
@@ -33,6 +33,18 @@ def range_slider(question)
...
+      range_body = tag(:input,
+                       type: 'range',
+                       id: idify(question[:id]),
+                       name: answer_name(idify(question[:id])),
+                       min: minmax[:min].to_s,
+                       max: minmax[:max].to_s,
+                       step: question[:step] || 1,
+                       required: true,
+                       value: question[:value])
+    end
```
emerencia, line 47: Kan je al die properties niet gewoon in een hash zetten, en daar dan evt. dat value attribuut aan toevoegen?
En dan roep je `tag` aan met `:input` en dan die hash.
Dat scheelt een boel codeduplicatie.

emerencia (approved):
Verder LGTM

app/generators/range_generator.rb
emerencia:
```

  def range_slider(question)
    minmax = range_slider_minmax(question)
    question_options = {
      type: 'range',
      id: idify(question[:id]),
      name: answer_name(idify(question[:id])),
      min: minmax[:min].to_s,
      max: minmax[:max].to_s,
      step: question[:step] || 1,
      required: true
    }
    question_options[:value] = question[:value] if question[:value].present?
    range_body = tag(:input, question_options)
    tag.p(range_body,
          class: 'range-field')
  end
```

PR #1704, opened 2020-09-03
[34286115] Added fancy time picker

frbl:
# Description of feature
...

# Checklist


emerencia:
Don't merge this yet because we'd need to adjust the base-platform in two locations:

Line 151 of app/javascript/components/analysis/dashboards/questionnaire/Utils.jsx:

rubymine64_2020-10-07_18-18-06

And these lines in u_can_act_brewer.rb:

rubymine64_2020-10-07_18-19-19

Also, the above lines should be changed to how the date is parsed in the JavaScript version (with the fallback check for uren (for compatibility with older questionnaire responses in case of a reroasting) and the zero padding before the time parsing).

emerencia:
@frbl Can we also use this for the date inputs?

PR #1742, opened 2020-10-14
Range updates & scheduling fixes

emerencia:
## In this PR:

- Enables ticks on sliders.
- Fixes OTR bugs with scheduled protocols.
- Fixes that protocols could be started in the past, resulting in no responses being scheduled.
- Added translation for Sunday questionnaire

Large screen size:

chrome_2020-10-19_19-58-30

Small screen size:

chrome_2020-10-19_19-57-54


app/controllers/api/v1/basic_auth_api/protocol_subscriptions_controller.rb
```diff
@@ -49,13 +49,17 @@ def set_protocol_subscription
...
-          Time.zone.parse(protocol_subscription_create_params[:start_date])
+          # The start date cannot be in the past.
+          [Time.zone.parse(protocol_subscription_create_params[:start_date]), Time.zone.now].max
         end
 
         def end_date
           return nil if protocol_subscription_create_params[:end_date].blank?
 
-          Time.zone.parse(protocol_subscription_create_params[:end_date])
+          # The duration between start and end date should be at least one hour.
```
frbl, line 59: why is this? Can you add a comment why it should be at least 1 hour?

frbl (approved):
lgtm

app/controllers/api/v1/basic_auth_api/protocol_subscriptions_controller.rb
emerencia: So people have at least one hour to fill out the measurement.

emerencia: I added a comment.

PR #1782, opened 2020-11-25
Implement localized date and time control that is backwards compatible for training log

emerencia:


app/generators/date_and_time_generator.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  def mydate_and_time_field(question)
+    body = mydate_and_time_tag(question)
+    body = tag.div(body, class: 'col s12 m6')
+    tag.div(body, class: 'row')
+  end
+
+  # rubocop:disable Metrics/AbcSize
+  def mydate_and_time_tag(question)
+    locale = (question[:locale].presence || Rails.application.config.i18n.default_locale).to_sym
```
frbl, line 22: just double checking, that `.presence` is correct here?

app/javascript/components/DateAndTimePicker.jsx
```diff
@@ -0,0 +1,64 @@
...
+        {...dateAndTimePickerDefaultOptions}
+        selected={startDate}
+        onChange={date => setStartDate(date)}
+        minDate={calculateMin()}
+        maxDate={calculateMax()}
+        required={props.required}
+        showTimeSelect
+        timeIntervals={15}
+        locale={(props.locale && props.locale === 'nl') ? nl : enUS}
+        timeCaption={(props.locale && props.locale === 'nl') ? 'Tijd' : 'Time'}
```
frbl, line 49: not a big fan of this, but I think its fine, especially since we don't have i18n right now her, right?

frbl (approved):
lgtm

app/javascript/components/DateAndTimePicker.jsx
emerencia: I think we do have i18n here. I changed it to an i18n string where we just set the locale parameter. It works.

app/generators/date_and_time_generator.rb
emerencia: Yep. It `question[:locale].presence` is equivalent to `question[:locale] if question[:locale].present?`.

It was actually rubocop that automatically changed it into this for me.

PR #1838, opened 2021-01-18
[36011120] Also export canceled and completed protocols, and export the state

emerencia:
In this pulll request:

- Fix to do the cleanup of reponses in prot sub cancel! method in a delayed job.
- Fix to not schedule more than 500 responses in the future for a single prot sub.
- Export canceled and completed protocol subscriptions, and export their state through the API.
- Allow updating a protocol subscription (just the end date for now).


app/controllers/api/v1/basic_auth_api/protocol_subscriptions_controller.rb
```diff
@@ -5,9 +5,9 @@ module V1
     module BasicAuthApi
       class ProtocolSubscriptionsController < BasicAuthApiController
         before_action :set_person, only: %i[create destroy_delegated_protocol_subscriptions]
-        before_action :set_external_identifier, only: %i[delegated_protocol_subscriptions destroy
+        before_action :set_external_identifier, only: %i[delegated_protocol_subscriptions destroy update
                                                          destroy_delegated_protocol_subscriptions]
-        before_action :set_protocol_subscription, only: %i[destroy]
+        before_action :set_protocol_subscription, only: %i[destroy update]
```
frbl, line 10: also update swagger? If we have it

app/controllers/api/v1/basic_auth_api/protocol_subscriptions_controller.rb
```diff
@@ -27,7 +27,7 @@ def create
         end
 
         def delegated_protocol_subscriptions
```
frbl, line 29: this is quite the change tbh if you're relying on this api to be stable. We should probably also document this in the future somewhere

app/controllers/api/v1/basic_auth_api/protocol_subscriptions_controller.rb
```diff
@@ -43,11 +43,25 @@ def destroy
           destroyed
         end
 
+        # This updates the protocol subscription. Only works if the external_identifier is given.
+        # We rescheduel the responses with a date that is in the future, for the same reasoning
+        # as is explained in reschedule_responses.rb.
+        def update
+          res = @protocol_subscription.update(protocol_subscription_update_params)
+          if res
```
frbl, line 51: can't you do `return unprocessable_entity(@protocol_subscription.errors) if res.nil?`

app/jobs/destroy_future_responses_job.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+class DestroyFutureResponsesJob < ApplicationJob
+  queue_as :default
+
+  def perform(protocol_subscription_id)
+    protocol_subscription = ProtocolSubscription.find_by(id: protocol_subscription_id)
+
+    # It's possible that hte protocol subscription was destroyed,
```
frbl, line 9:
```suggestion
    # It's possible that the protocol subscription was destroyed,
```

app/models/measurement.rb
```diff
@@ -5,6 +5,15 @@ class Measurement < ApplicationRecord
   WEEKDAYS = %w[sunday monday tuesday wednesday thursday friday saturday].freeze
   MIN_PRIORITY = -10_000
   MAX_PRIORITY = 10_000
+  # How many responses will at most be scheduled at once for a single protocol subscription
+  # by the RescheduleResponses job. Since this job is ran every day, as long as this number is
+  # at least as large as the number of responses per day, it will be called again before we run
+  # out of responses, so it is all fine. The only issue here is that we use the response count
+  # to give people a measure of "completion", which we now have to add a note to that says
```
frbl, line 12: ah check yeah good point

app/models/measurement.rb
```diff
@@ -5,6 +5,15 @@ class Measurement < ApplicationRecord
...
+  # if there are more than 500 responses to be filled out by the user, this completion count
+  # or percentage can be off. (Since it just looks at how many responses were created for
+  # a protocol subscription.)
```
frbl, line 15: aldernative would be to calculate the number of responses instead of actually scheduling them?

frbl (approved):
lGtM

app/controllers/api/v1/basic_auth_api/protocol_subscriptions_controller.rb
emerencia: Done!

app/controllers/api/v1/basic_auth_api/protocol_subscriptions_controller.rb
emerencia: Yep. Here's the thing: it's only used for SDV currently. That the only reason that I was able to make this change.

app/models/measurement.rb
emerencia: The problem is that the protocol completion is requested frequently and should be fast. So what if someone schedules a protocol subscription for the next 100 years, do we really want to return a possibly unlimited size array with protocol_completion entries that takes a long time to construct (even if we're not making responses) and uses a lot of memory?

Or just say that it doesn't work right in these cases.

app/controllers/api/v1/basic_auth_api/protocol_subscriptions_controller.rb
emerencia: Sure can.

emerencia: Fixed.

PR #1857, opened 2021-02-03
Invitation link fixes

emerencia:
- Fix invitation set bug with multiple responses in an invitationset the link would no longer work for non mentors if the first response in the invitation set was filled out. (Fixed SDV bug with Sunday questionnaire reminders).
- Adds option for u-can-feel to allow login through UUID (to hopefully fix Outlook/Edge/... redirect/cookie setting issues). This new setting defaults to false for all projects except u-can-feel.

### DIFF
--- a/app/controllers/token_authentication_controller.rb
+++ b/app/controllers/token_authentication_controller.rb
@@ -41,14 +41,6 @@ class TokenAuthenticationController < ApplicationController
     store_person_cookie(identifier_param)
   end
 
-  def store_person_cookie(identifier)
-    cookie = { PERSON_ID_COOKIE => identifier }
-    CookieJar.set_or_update_cookie(cookies.signed, cookie)
-    # Remove the JWTAuthenticator cookie if set, so we log out any other sessions on this browser.
-    CookieJar.delete_cookie(cookies.signed, JWT_TOKEN_COOKIE)
-    store_verification_cookie
-  end
-
   def identifier_param
     identifier = questionnaire_params[:q]
     identifier[0...Person::IDENTIFIER_LENGTH] if identifier
--- a/app/controllers/token_authentication_controller.rb
+++ b/app/controllers/token_authentication_controller.rb
@@ -3,29 +3,39 @@
 class TokenAuthenticationController < ApplicationController
   before_action :check_params
   before_action :check_invitation_token
+  before_action :set_all_responses
 
   RESPONSE_ID_COOKIE = :response_id
   JWT_TOKEN_COOKIE = :jwt_token
   PERSON_ID_COOKIE = :person_id
 
   def show
-    responses = InvitationToken.find_attached_responses(questionnaire_params[:q])
-    if responses.blank?
-      render(status: :not_found, html: 'Deze link is niet (meer) geldig.', layout: 'application')
-      return
-    end
-    current_person = responses.first.person
-    if current_person.mentor? && !responses.first.protocol_subscription.for_myself?
-      # If we are a mentor, we want to follow the full logic of the next page finder
+    current_person = @all_responses.first.person
+    only_non_completed_responses = InvitationToken.find_attached_responses(questionnaire_params[:q], true)
+    if current_person.mentor? &&
+       (!@all_responses.first.protocol_subscription.for_myself? || only_non_completed_responses.blank?)
+      # If we are a mentor, and the first response was already filled out or there is no response to fill out,
+      # redirect to the mentor dashboard and/or follow the logic of the next page finder.
       redirect_to questionnaire_index_path
+    elsif only_non_completed_responses.blank?
+      # If we are not a mentor or the prot sub is for ourselves, then we want to redirect to the response,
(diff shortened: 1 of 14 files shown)


config/settings.yml
```diff
@@ -13,6 +13,7 @@ default: &default
     allow_distribution_export: false
     realtime_distributions: true
     google_analytics: true
+    allow_response_uuid_login: false
```
frbl, line 16: do we have documentation somewhere for these options? Perhaps also update it there

spec/controllers/questionnaire_controller_spec.rb
```diff
@@ -121,6 +121,56 @@
         Timecop.return
       end
     end
+
```
frbl, line 124: not sure if we have a swagger spec for this route, but if we do also update that one.

frbl (approved):
lgtm

app/controllers/token_authentication_controller.rb
```diff
@@ -3,30 +3,40 @@
...
-    if current_person.mentor? && !responses.first.protocol_subscription.for_myself?
-      # If we are a mentor, we want to follow the full logic of the next page finder
+    current_person = @all_responses.first.person
+    only_non_completed_responses = InvitationToken.find_attached_responses(questionnaire_params[:q], true)
+    if current_person.mentor? &&
+       (!@all_responses.first.protocol_subscription.for_myself? || only_non_completed_responses.blank?)
+      # If we are a mentor, and the first response was already filled out or there is no response to fill out,
+      # redirect to the mentor dashboard and/or follow the logic of the next page finder.
       redirect_to questionnaire_index_path
+    elsif only_non_completed_responses.blank?
```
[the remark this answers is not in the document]
emerencia: A redirect_to does not return from the function itself. No function can. Only a ruby `return` keyword does that. That's why we do all those `render && return` statements.

app/helpers/application_helper.rb
```diff
@@ -150,6 +164,26 @@ def current_user_from_header
...
+    return nil unless Rails.application.config.settings.feature_toggles.allow_response_uuid_login
+
+    # Works on both the questionnaire and the questionnaire/preference urls
+    uuid_match = %r{\A/questionnaire/(?:preference/)?([a-z0-9-]+)/?\z}.match(request&.path)
+    return nil unless uuid_match.present?
+
+    # The response must be opened and cannot be completed already
+    # (this is so that once a response has been completed, that link cannot
+    #  be used to log in for that same user again)
+    response = Response.opened.find_by(uuid: uuid_match[1])
```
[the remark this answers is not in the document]
emerencia: yes because the `(?:)` introduces a regex backreference that is not included in the match list (as opposed to a `()` match that is included). I'll add a comment.

config/settings.yml
emerencia: Good idea. I'll add a section on feature toggles to the readme.

spec/controllers/questionnaire_controller_spec.rb
emerencia: Thanks, I'll run the swagger thing. We should really have a circleCI job that checks if the swagger spec is up-to-date.

app/controllers/token_authentication_controller.rb
```diff
@@ -3,30 +3,40 @@
...
+      # If we have an open response to redirect to, give priority to the response from
+      # the set we actually clicked on (i.e., the first in the invitation set that is still open).
+      redirect_to preference_questionnaire_index_path(uuid: only_non_completed_responses.first.uuid)
     end
   end
 
   private
 
+  def set_all_responses
+    @all_responses = InvitationToken.find_attached_responses(questionnaire_params[:q], false)
```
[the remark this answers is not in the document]
emerencia: Ok

app/controllers/token_authentication_controller.rb
```diff
@@ -3,30 +3,40 @@
...
+      # If we are not a mentor or the prot sub is for ourselves, then we want to redirect to the response,
+      # but only if it is not yet completed. Otherwise give an error.
+      render(status: :not_found, html: 'Deze link is niet (meer) geldig.', layout: 'application')
     else
-      # If we are a student, we want to give priority to the response from
-      # the set we actually clicked on
-      redirect_to preference_questionnaire_index_path(uuid: responses.first.uuid)
+      # If we have an open response to redirect to, give priority to the response from
+      # the set we actually clicked on (i.e., the first in the invitation set that is still open).
+      redirect_to preference_questionnaire_index_path(uuid: only_non_completed_responses.first.uuid)
```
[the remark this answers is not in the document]
emerencia: Yep you're right it was confusing. I refactored it now so that it does the same, but it is much more clear and concise.

spec/controllers/questionnaire_controller_spec.rb
emerencia: Fixed.

config/settings.yml
emerencia: Done.

PR #1886, opened 2021-02-17
Upgrade Rails to 6.0.3.5

emerencia:


config/initializers/lograge.rb
```diff
@@ -2,11 +2,19 @@
...
+  # In some test specs, where we test JWT login without specifying a token,
+  # current_user raises an error, hence this fallback.
   config.lograge.custom_payload do |controller|
-    {
-      ip: controller.request.ip,
-      user_id: controller.current_user.try(:id)
-    }
+    begin
+      {
+        ip: controller.request.ip,
```
frbl, line 10: do we really need to log IPs? Id rather not tbh

frbl (approved):
lgtm

config/initializers/lograge.rb
emerencia: Ok, I'll remove it.

emerencia: Fixed.

PR #1904, opened 2021-03-03
[38140136] Fix email/mobile phone uniqueness when setting to empty values

emerencia:
Fixes https://freedcamp.com/view/2354427/tasks/panel/task/38140136

### DIFF
--- a/app/controllers/api/v1/cookie_and_jwt_api/person_controller.rb
+++ b/app/controllers/api/v1/cookie_and_jwt_api/person_controller.rb
@@ -28,12 +28,24 @@ module Api
 
         private
 
+        # If we are sent an empty string as an email, set it to nil instead. Because we do
+        # uniqueness checking on the email, having multiple people with an empty string as
+        # email is not allowed, but having multiple people with a nil email is allowed.
+        # The same applies to mobile phone numbers.
+        def overwritten_params
+          result = {}
+          %i[email mobile_phone].each do |param|
+            result[param] = nil if person_params.key?(param) && !person_params[param].present?
+          end
+          result
+        end
+
         def update_person
-          timestamp = person_params[:timestamp] and person_params[:timestamp].to_datetime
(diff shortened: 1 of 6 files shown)


app/controllers/api/v1/cookie_and_jwt_api/person_controller.rb
```diff
@@ -28,12 +28,24 @@ def destroy
...
+
         def update_person
-          timestamp = person_params[:timestamp] and person_params[:timestamp].to_datetime
+          timestamp = person_params[:timestamp].present? ? Time.zone.parse(person_params[:timestamp]) : nil
           # no timestamp: request is synchronous
           # if timestamp: request is async, so check if it is the most recent one
-          if timestamp.nil? || (timestamp > current_user.updated_at)
-            return current_user.update(person_params.except(:timestamp))
+          if timestamp.blank? || timestamp > current_user.updated_at
+            return current_user.update(person_params.except(:timestamp).merge(overwritten_params))
```
frbl, line 48: do we only want to do this for an update? Or should we also do this for a create?

frbl (approved):
lgtm

app/controllers/api/v1/cookie_and_jwt_api/person_controller.rb
emerencia: That's a very good point. I'll see if it affects create as well.

emerencia: Ok I looked this up and the users are created as part of the auth_user.rb/JWT authentication flow, which eventually calls `CreateAnonymousUser`, which always initializes the users with a nil email and mobile_phone number.

emerencia: I checked all the other people_controllers and person_controllers for methods that update or create a person object. I found one more controller where it is possible to update a person's email and/or mobile_phone, I'll fix that one also.

emerencia: Fixed.

PR #1905, opened 2021-03-03
[38233767] check logic for sending invitations again when sending reminders

emerencia:
Fixes https://freedcamp.com/view/2354427/tasks/panel/task/38233767

### DIFF
--- a/app/jobs/send_invitations_job.rb
+++ b/app/jobs/send_invitations_job.rb
@@ -5,7 +5,15 @@ class SendInvitationsJob < ApplicationJob
 
   def perform(invitation_set)
     invitation_set.reload
+    # Since this can be the reminder notification, which is sent hours later,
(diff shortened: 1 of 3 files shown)


frbl (approved):
lgtm, I do wonder if we want to make a wrapper for those 3 calls that you're doing

emerencia:
> lgtm, I do wonder if we want to make a wrapper for those 3 calls that you're doing

I don't think we should, because it is only part of the logic that we're checking. It doesn't check whether a response is opened and not expired or completed, but that is checked as a scope on the previous line. So this specific selection of methods only applies when you already checked that other scope.

So any refactoring into a new method would have to add the checks from the scope to be complete, and then we have a much more inefficient query here, with checking the scoped checks twice.

So I'll leave it like this for now. Also because it doesn't occur anywhere else in the code.

PR #2001, opened 2021-05-14
Draft: Update webpacker

emerencia:


emerencia:
Will have to revisit this since we upgraded to rails 6 and whatnot.

PR #2004, opened 2021-05-26
[Hotfix] logging for missing phone numbers

frbl:
# Description of feature
...

# Checklist


m-rios (approved):

emerencia:
This doesn't need to call appsignal.

It could very well happen that someone has set their email or mobile_phone number to nil somewhere in the 8 hours waiting period between invites, and it shouldn't give an error.

You can see that in email_invitation.rb, we do handle this error correctly. I'll apply the fix in a different branch.

PR #2012, opened 2021-06-23
Draft: [40588882] Add support for supporting different questionnaire renderers Complete Task

Kranex:
# Description of feature
Adds a new_renderer parameter to the `/questionnaire` route which, if set to `'true'`, will use react to render the questionnaire rather than the current method.

The react renderer is imported from an external npm library that can be found here: https://www.npmjs.com/package/sdv-questionnaire-test-renderer.

# Checklist

### DIFF
--- a/app/controllers/questionnaire_controller.rb
+++ b/app/controllers/questionnaire_controller.rb
@@ -288,16 +288,23 @@ class QuestionnaireController < ApplicationController
   end
 
   def set_questionnaire_content
-    @content = QuestionnaireGenerator.new.generate_questionnaire(
-      response_id: @response.id,
-      content: @response.measurement.questionnaire.content,
-      title: @response.measurement.questionnaire.title,
-      submit_text: (@response.person.locale == 'en' ? 'Save' : 'Opslaan'),
-      action: '/',
-      unsubscribe_url: @response.unsubscribe_url,
-      locale: @response.person.locale,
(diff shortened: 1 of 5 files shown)


app/controllers/questionnaire_controller.rb
```diff
@@ -287,17 +287,24 @@ def set_protocol_and_subscription
...
-      params: default_questionnaire_params
-    )
+  def set_questionnaire_content # rubocop:disable Metrics/AbcSize
+    use_new_renderer = questionnaire_params[:new_renderer]
+
+    if use_new_renderer.present? && use_new_renderer.casecmp('true').zero?
+      @renderer = :new
+      @content = @response.measurement.questionnaire.content
+    else
+      @content = QuestionnaireGenerator.new.generate_questionnaire(
```
emerencia, line 297: Please also define @renderer in the else body

app/views/questionnaire/show.html.haml
```diff
@@ -1,2 +1,5 @@
 = render partial: 'shared/header'
-= @content
+- if @renderer == :new
+  = react_component('Questionnaire', props= {content: @content})
```
emerencia, line 3:
```suggestion
  = react_component('Questionnaire', content: @content)
```

emerencia:
Can you add some specs? Like that it calls the new renderer?

PR #2041, opened 2021-06-28
[40588927] Update daily questionnaire with new text

emerencia:
Fixes https://freedcamp.com/view/2354427/tasks/panel/task/40588927


frbl (approved):
lgtm

PR #2054, opened 2021-07-09
Add days question type

emerencia:


config/locales/en.yml
```diff
@@ -1,3 +1,243 @@
 en:
```
m-rios, line 1: why didn't these translations exist before?

app/generators/days_generator.rb
```diff
@@ -0,0 +1,80 @@
...
+    from_days_ago = question[:from_days_ago]
+    date_format = question[:date_format].presence || DATE_FORMAT
+    from_days_ago.downto(today) do |days_ago|
+      date = days_ago.days.ago
+      next if [6, 0].include?(date.wday) && question[:exclude_weekends].present?
+
+      formatted_date = I18n.l(date, locale: question[:locale], format: date_format)
+      option = {
+        title: formatted_date,
+        value: I18n.l(date, locale: :en, format: '%Y-%m-%d')
```
m-rios, line 38: why don't you want to use `date_format` here? Should this special format be in a constant?

m-rios (approved):

config/locales/en.yml
emerencia: Because we never had to use Rails time localization for anything. The date control was front-end only. This is backend.

app/generators/days_generator.rb
emerencia: Good point. Let me fix it.

emerencia: Sorry, now I remember again. This is the value that is used as form value, i.e., the user never sees this value. It is merely what will be sent to the backend. And for this I wanted something that was 1) obviously always the same, no matter the user's locale (that's why it is always in English), and 2) something that would be easy to parse, but was still custom made, so it's the date, optionally followed by AM or PM.

I'll move everything to constants and add an explanation.

emerencia: Fixed.

PR #2064, opened 2021-07-14
[40988616] Add OSTRC scores

emerencia:
Fixes https://freedcamp.com/view/2354427/tasks/panel/task/40988616

### DIFF
--- a/app/use_cases/calculate_scores.rb
+++ b/app/use_cases/calculate_scores.rb
@@ -111,6 +111,8 @@ class CalculateScores < ActiveInteraction::Base
     case score[:operation]
     when :average
       perform_operation_average(data)
+    when :sum
+      perform_operation_sum(data)
     else
       raise "unknown operation: #{score[:operation]}"
     end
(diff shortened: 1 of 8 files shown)


app/use_cases/calculate_scores.rb
```diff
@@ -123,6 +125,13 @@ def perform_operation_average(data)
     data.sum(0.0) / data.size
   end
 
+  def perform_operation_sum(data)
+    raise MyMissingDataError, 'trying to calculate the sum of an empty array' if data.size.zero?
+    return data[0] if data.size == 1 # no need to convert to float if we have just one integer
+
+    data.sum(0)
```
m-rios, line 132: do we need to convert to float otherwise? or is that taken care of by sum?

projects/sport-data-valley/seeds/questionnaire_helpers/complaints.rb
```diff
@@ -418,5 +422,23 @@ def all_complaint_options(with_otherwise = false)
       end
       result
     end
+
+    def all_complaint_scores(with_otherwise = false)
```
m-rios, line 426: shouldn't this be 'compliance'? (same for the function below)

m-rios (approved):

m-rios:
LGTM

app/use_cases/calculate_scores.rb
emerencia: Nope there is no need to convert to float here. If one of the values in a calculation is a Float, ruby will automatically convert the whole thing to a float. And otherwise we can just stay as an integer (which is the same as when rounding to 0 decimals).

projects/sport-data-valley/seeds/questionnaire_helpers/complaints.rb
emerencia: No, these are about complaints people have (as in physical ailments). So they are literally "complaint scores".

PR #2069, opened 2021-07-20
[40481557] Add umo project

emerencia:
Fixes https://freedcamp.com/view/2914531/tasks/panel/task/40481557


projects/umo/config/settings.yml
```diff
@@ -0,0 +1,19 @@
+# Beware! This file is public by definition
+default: &default
+  default_team_name: 'Demo-team'
```
m-rios, line 3: should this be UMO? or does it not matter?

m-rios (approved):

projects/umo/config/settings.yml
emerencia: We can change it once we make a team. It should correspond to an existing team name in seeds/teams. It's for when we use OTR responses, to add those anonymously created people to a team/organization.

PR #2080, opened 2021-07-28
[40986263] Preview protocols and allow to specify open_from_day with offsets

emerencia:
Fixes https://freedcamp.com/view/2354427/tasks/panel/task/40986263


app/helpers/protocol_helper.rb
```diff
@@ -0,0 +1,24 @@
...
+  def end_date(start_date_str, end_date_str)
+    return nil if end_date_str.blank?
+
+    # The duration between start and end date should be at least one hour.
+    # There is no specific reason for having a one hour minimum, it's an
+    # artibtrary amount of time that should be enough to fill out any questionnaire.
+    # The limit is to make it more fool proof, i.e., there's no case in which
+    # you would want to start a protocol shorter than one hour (or at least
+    # if there is such a case, we don't allow it through this API).
+    minimum_end_date = TimeTools.increase_by_duration(start_date(start_date_str), 1.hour)
```
m-rios, line 20: should we create a constant for this? or is this 1 hour limit only used here?

m-rios (approved):

app/helpers/protocol_helper.rb
emerencia: You're right. No magic numbers! I'll create a constant.

emerencia: Fixed.

PR #2090, opened 2021-08-05
[41426178] Add protocol completion data to push subscription pushed response data

emerencia:
Fixes https://freedcamp.com/view/2354427/tasks/panel/task/41426178


frbl:
LGTM

PR #2150, opened 2021-09-18
[41491814] Fix questionnaire redirect logic

emerencia:
Fixes https://freedcamp.com/view/2354427/tasks/panel/task/41491814

Extends Measurements with a new property `only_redirect_if_nothing_else_ready`. This property causes the questionnaire engine to redirect to the next response if one is ready even when a redirect url is defined for the measurement.


frbl (approved):
lgtm

PR #2160, opened 2021-09-20
[42304854] Remove future responses from protocol completion and add properties

emerencia:
Add measurement ID and protocol subscription ID to response serializer to allow them to be identified better with the protocol_completions.

Don't include future responses in the protocol completion when used in the response serializer. This is because it takes up a LOT of bytes for long protocols and we don't need this information here.

Fixes https://freedcamp.com/view/2354427/tasks/panel/task/42304854


app/serializers/api/response_serializer.rb
```diff
@@ -19,9 +19,15 @@ def questionnaire
...
-      return object.protocol_subscription.protocol_completion if instance_options[:with_protocol_completion]
+      return protocol_completion_without_future if instance_options[:with_protocol_completion]
 
       []
     end
+
+    private
+
+    def protocol_completion_without_future
+      object.protocol_subscription.protocol_completion.reject { |entry| entry[:future].present? }
```
frbl, line 30: move this to protocol subscription instead? Or well, it is also just representation logic, so the serializer might be a good place.

frbl (approved):
lgtm

app/serializers/api/response_serializer.rb
emerencia: No because there are still places where we DO want the entire protocol completion, for example when we show a progress bar, in order to calculate how much percentage someone has completed of a diary study, we do need the number of future responses. I'll add this in a comment to clarify.

emerencia: Fixed.

PR #2189, opened 2021-10-04
[40455442] Add umo protocols and questionnaires

emerencia:
Fixes https://freedcamp.com/view/2914531/tasks/panel/task/40455442


projects/new/seeds/questionnaires/new.rb
```diff
@@ -1,11 +1,11 @@
 # frozen_string_literal: true
 
-# db_title = 'Demo vragenlijst' # Dagboekvragenlijst moet geen titel hebben alleen een logo
+# db_title = '' # Shown to users as a big text, but cannot be localized, so often left empty.
```
m-rios, line 3: this is all commented, can it be deleted just like the other seeds?

app/models/questionnaire.rb
```diff
@@ -16,7 +16,7 @@ class Questionnaire < ApplicationRecord
 
   validates :name, presence: true, uniqueness: true
   validates :content, presence: true
-  validates :key, presence: true, uniqueness: true, format: { with: /\A[a-z]+[a-z_0-9]*\Z/ }
```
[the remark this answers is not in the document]
emerencia: Correct. Because some of them were named `01_demographics` and stuff. 

The reason for not allowing a key to start with a number is that converting a key to a symbol is then no longer possible. Symbols in Ruby can't start with a digit because of syntax reasons.
```ruby
[1] pry(main)> :01_demographics
(eval):2: trailing `_' in number
:01_demographics
   ^
SyntaxError: unexpected integer literal, expecting literal content or terminator or tSTRING_DBEG or tSTRING_DVAR
```

However, I don't believe that we symbolize the key of a questionnaire anywhere. Feel free to double check this.

projects/new/seeds/questionnaires/new.rb
emerencia: So this is the "new" project. Which is when you call the `deployment:create_project` rake task, it essentially copies this whole directory so that you have something to start from with some example files. 

Typically you would delete these files and put your own instead, but we also use it as like a hint for new projects so they know what to think about when making seeds (or to remind ourselves). For this reason, I'd like to leave the commenting in the seeds in the `new` project folder as verbose as possible.

Here I just changed the comments to English because they were in Dutch before, and I updated the syntax to reflect that for most questionnaires these days, the questionnaire key is the same as the filename.

PR #2241, opened 2021-11-01
Add support for shows_questions and hides_questions attributes for dropdowns and upgrade webpacker

emerencia:


app/assets/javascripts/questionnaire_logic.js.erb
```diff
@@ -64,6 +64,19 @@ function toggle_shown_questions() {
   }
 }
 
+function toggle_shown_questions_for_dropdown () {
```
frbl, line 67: why didn't you structure this the same way as the `toggle_hidden_questions_for_dropdown` function? Its the same right? except for the `!`

app/assets/javascripts/questionnaire_logic.js.erb
```diff
@@ -276,8 +299,10 @@ function questionnaireLoaded() {
   const array_of_dom_elements = document.querySelectorAll('input[type=range]')
   M.Range.init(array_of_dom_elements)
   $('input[type=radio],input[type=checkbox]').change(toggle_shown_questions).change(toggle_otherwise_field).change(toggle_hidden_questions);
+  $('select').change(toggle_shown_questions_for_dropdown).change(toggle_hidden_questions_for_dropdown);
   $(document).on('touchstart mousedown click change', '.notchanged input[type=range]', removeNotChanged);
   $('input[type=radio][data-shows-questions],input[type=checkbox][data-shows-questions]').each(toggle_shown_questions);
+  $('option[data-shows-questions]').each(toggle_shown_questions_for_dropdown);
```
frbl, line 305: you only need to do this for 'shows'?

package.json
```diff
@@ -26,19 +26,21 @@
...
     "@babel/preset-env": "^7.5.5",
     "babel-jest": "^24.9.0",
     "babel-plugin-transform-class-properties": "^6.24.1",
     "babel-plugin-transform-decorators-legacy": "^1.3.5",
+    "dotenv-webpack": "^7.0.3",
     "enzyme": "^3.10.0",
     "enzyme-adapter-react-16": "^1.14.0",
     "jest": "^26.6.3",
-    "webpack-dev-server": "^3.8.0"
+    "webpack-dev-server": "^3"
```
frbl, line 43: is this actually a downgrade?

frbl (approved):
lgtm

app/assets/javascripts/questionnaire_logic.js.erb
emerencia: Good point, I'll fix it.

emerencia: Actually, there's 3 differences.

- The `.find('option[data-hides-questions]')` instead of data-shows-questions
- The `.data('hides-questions')` instead of shows-questions
- and then the !this_is_checked instead of this_is_checked

I don't see a clean way to use one function for this so I'll keep them separate for now.

This code is only temporary anyways, until we have the react implementation of all of this.

package.json
emerencia: So this is what came out when i ran the rails-webpack generator.
I checked the yarn.lock file, and there it's using version 3.11.2 now, so that is fine.

I'll change it to 3.11.2

emerencia: Fixed.

app/assets/javascripts/questionnaire_logic.js.erb
emerencia: So I copied this from how we handle it for radios and checkboxes, which is strictly not correct.

What the toggle_shown_questions* functions actually do is check if a value is selected and if it is, remove the hidden class from any objects that it might show according to the shows_questions notation.

We should technically be able to do the same for the toggle_hidden_questions* functions, but a comment in the code says that if you do that, it breaks everything.

The reason is that we do not check for visibility, i.e., we remove the hidden class even from questions that are themselves not shown because of other shows_questions options. And this is because we change all their visibility in one step, instead of going over everything recursively. So we can't just ignore currently hidden objects and not show their 

The only reason that this works is that we assume that the initial visibility of the questions in the questionnaire definitions is such that it does not need changing. I.e., if a question is shown by the shows_question attribute of an earlier question, that question will have hidden: true in the questionnaire definition.

So then by that logic, we do not need to check the visibility for the shows_questions either, but I think there was a reason that we do it here. Like that in the interactive parsing route, we don't have a ready event so we call this function when the dom gets updated? I'm not sure. Either way this again is temporary code and I don't want to spend too much time on it if it works because it will all be replaced by react logic soon.

PR #2242, opened 2021-11-03
Protocols for Fall 2021 measurement

everenn:
New protocols for the three different groups participating in the fall 2021 measurement: people who participated in the spring (who have already consented), people who joined but did not participate in the spring, and people who joined in the fall.


emerencia (approved):
LGTM

PR #2280, opened 2021-11-22
chore: add seeds for the training log questionnaire

m-rios:
# Description of feature
...

# Checklist


lib/tasks/manual_seed.rake
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+# This task allows to manually execute a custom seed file. To run the task call `rails db:seed:file_name` where
+# file_name is that of the seed you want to execute (without the extension)
+namespace :db do
+  namespace :seed do
+    Dir[Rails.root.join('projects', 'sport-data-valley', 'manual_seeds', '*.rb')].each do |filename|
```
emerencia, line 7:
```suggestion
    Dir[Rails.root.join('projects', ENV['PROJECT_NAME'], 'manual_seeds', '*.rb')].each do |filename|
```

projects/sport-data-valley/manual_seeds/rheuma_responses.rb
```diff
@@ -0,0 +1,90 @@
+# frozen_string_literal: true
+
+require_relative 'helpers/people_helper'
+
+srand(123)
+
+START_DATE = Time.zone.parse('2021-11-24').beginning_of_day
```
emerencia, line 7: Can we do something dynamic? Like 3.weeks ago, and then if the end_date is 3.weeks ago + 3 weeks = today. So that when we run the script we always generate a bunch of recent responses? So that it looks good in the dashboard graphs.

README.md
```diff
@@ -176,6 +176,19 @@ In order to run the Capybara specs of the VSV project, you need to install the c
   brew install chromedriver
 ```
 
+### Development seeds
+At some point you might need pre-existing questionnaire responses in order to test the applications that integrate 
+with the questionnaire engine. You can pre-seed the database with fake data using manual seeds. At the moment this 
+feature is only available for the `sport-data-valley` project. To run the manual seeds, simply call:
```
emerencia, line 182: I think we can make this available for all projects, see my comment below.

projects/sport-data-valley/manual_seeds/helpers/people_helper.rb
```diff
@@ -0,0 +1,21 @@
...
+  person ||= Person.new(email: email)
+
+  person.first_name = 'Voornaam'
+  person.last_name = 'Achternaam'
+  person.role = normal_role
+  person.account_active = true
+
+  person.save!
+  person
+end
```
emerencia, line 21: Missing newline at the end.

projects/sport-data-valley/manual_seeds/helpers/people_helper.rb
```diff
@@ -0,0 +1,21 @@
+def find_or_create_person(email)
```
emerencia, line 1: Can you put this in a module, and then just include the module from another module?

It's not very Ruby-like (and error prone) to have all these things living in the global namespace.

projects/sport-data-valley/manual_seeds/rheuma_responses.rb
```diff
@@ -0,0 +1,90 @@
...
+(1..N_PARTICIPANTS).each do |idx|
+  person = find_or_create_person("questionnaire_participant_#{idx}@researchable.nl")
+
+  date = START_DATE
+  while date <= END_DATE
+    create_rheuma_responses(person, date)
+
+    date = date + 1.day
+  end
+end
```
emerencia, line 90: Unless it's like a script living in `/bin` I prefer ruby code to be always in classes or modules, rather than in the global namespace.

projects/sport-data-valley/manual_seeds/rheuma_responses.rb
```diff
@@ -0,0 +1,90 @@
...
+  response_params[:filled_out_by_id] = filled_out_by_id
+  response_params[:filled_out_for_id] = filled_out_for_id
+
+  response_params[:open_from] = date - 1.minute
+  response_params[:opened_at] = date
+  response_params[:completed_at] = date + 1.minute
+
+  content = {}
+  content[:v1] = range_0_10
+  content[:v2] = rand > 0.5 ? 'Ya' : 'Nee'
```
emerencia, line 63: It's "Ja" , not "Ya"

And more specifically, it's `"Ja<br />Zo ja, waar denkt u dat dit door kwam?"` (see my comment at https://github.com/compsy/u-can-act/pull/2303 )

emerencia: And then the answer is Ja... they also have to specify a reason in the text field (it is required):

```ruby
 "v2_ja_br_zo_ja_waar_denkt_u_dat_dit_door_kwam_text"=>"geen idee"
```

I'm guessing you don't use those in the dashboard rihgt now, but it's nice to have them in the response generator, so they generate accurate responses that we can maybe also use for testing the questionnaire detailed view.

projects/sport-data-valley/manual_seeds/rheuma_responses.rb
```diff
@@ -0,0 +1,90 @@
...
+
+  protsub = ProtocolSubscription.find_by(person: person, protocol: protocol)
+  protsub ||= ProtocolSubscription.new(person: person, protocol: protocol)
+
+  protsub.state = ProtocolSubscription::COMPLETED_STATE
+  protsub.start_date = Time.new(2016).in_time_zone
+  protsub.end_date = Time.new(2017).in_time_zone
+  protsub.save!
+
+  questionnaire_name = 'Dagelijkse vragenlijst reuma'
```
emerencia, line 37: So in my opinion, a questionnaire name can always change, but its key stays the same always.

I know that we violate this assumption in some SDV seeds. And for this reason both name and key are unique, so in practice it doesn't matter which one you use to look up.

But maybe you can change this lookup to be by key instead (`daily_questionnaire_rheumatism`).

projects/sport-data-valley/manual_seeds/rheuma_responses.rb
```diff
@@ -0,0 +1,90 @@
...
+  response_params[:measurement_id] = measurement_id
+  response_params[:filled_out_by_id] = filled_out_by_id
+  response_params[:filled_out_for_id] = filled_out_for_id
+
+  response_params[:open_from] = date - 1.minute
+  response_params[:opened_at] = date
+  response_params[:completed_at] = date + 1.minute
+
+  content = {}
+  content[:v1] = range_0_10
```
emerencia, line 62: Btw I think this generates a number, but it's important to note that what is stored in the DB is always a string.

e.g.:
```ruby
{"locale"=>"nl",
 "v1"=>"5.8",
 "v2"=>"Ja<br />Zo ja, waar denkt u dat dit door kwam?",
 "v2_ja_br_zo_ja_waar_denkt_u_dat_dit_door_kwam_text"=>"geen idee",
 "v3"=>"2.8",
 "v4"=>"7.4",
 "v5"=>"4",
 "v6"=>"2",
 "v7"=>"5.5",
 "v8"=>"7",
 "v1_timing"=>"37793",
 "v2_nee_timing"=>"1648",
 "v3_timing"=>"1132",
 "v4_timing"=>"1046",
 "v5_timing"=>"1558",
 "v6_timing"=>"1811",
 "v7_timing"=>"1257",
 "v8_timing"=>"3654",
 "v2_ja_br_zo_ja_waar_denkt_u_dat_dit_door_kwam_timing"=>"82",
 "v2_ja_br_zo_ja_waar_denkt_u_dat_dit_door_kwam_text_timing"=>"3961"}
```

So for this to be accurate all the numbers should be strings I think.

And don't mind the `_timing` values. You don't have to add those. We don't use them for anything currently.

projects/sport-data-valley/manual_seeds/training_responses.rb
```diff
@@ -0,0 +1,248 @@
+# frozen_string_literal: true
```
emerencia, line 1: Again I would prefer this to be in a class or a module.

You can do something basic, like

```ruby
srand(123)
require_relative 'helpers/people_helper' # Make sure PeopleHelper is defined in our scope

class TrainingResponsesSeeds
  include PeopleHelper
  class << self
     def run!
        # do whatever
     end

    private
    
    # all your private methods here

  end
end

# Then as final line in the file you have this, so that it runs when you load the file (just make sure it's not in an autoloaded path (which I think is fine currently):
TrainingResponsesSeeds.run!
```

And then we keep a very clean namespace even when running seeds, but we can still load seeds on demand that are not autoloaded everytime we start the application.

projects/sport-data-valley/manual_seeds/training_responses.rb
```diff
@@ -0,0 +1,248 @@
...
+require_relative 'helpers/people_helper'
+srand(123)
+
+START_DATE = Time.zone.parse('2021-08-01').beginning_of_day
```
emerencia, line 5: I would prefer relative dates in these kind of scripts, so that you can call them a year later and still have recent responses generated without any editing of the file.

projects/sport-data-valley/manual_seeds/training_responses.rb
```diff
@@ -0,0 +1,248 @@
...
+  puts "Creating responses for #{protocol_name}"
+  protocol = Protocol.find_by(name: protocol_name)
+  raise "Error: protocol named #{protocol_name} not found" unless protocol
+
+  protsub = ProtocolSubscription.find_by(person: person, protocol: protocol)
+  protsub ||= ProtocolSubscription.new(person: person, protocol: protocol)
+
+  protsub.state = ProtocolSubscription::COMPLETED_STATE
+  protsub.start_date = Time.new(2016).in_time_zone
+  protsub.end_date = Time.new(2017).in_time_zone
```
emerencia, line 85: Here also the i would prefer relative dates

projects/sport-data-valley/manual_seeds/training_responses.rb
```diff
@@ -0,0 +1,248 @@
...
+
+  protocol = Protocol.find_by(name: protocol_name)
+  raise "Error: protocol named #{protocol_name} not found" unless protocol
+
+  protsub = ProtocolSubscription.find_by(person: person, protocol: protocol)
+  protsub ||= ProtocolSubscription.new(person: person, protocol: protocol)
+
+  protsub.state = ProtocolSubscription::COMPLETED_STATE
+  protsub.start_date = Time.new(2016).in_time_zone
+  protsub.end_date = Time.new(2017).in_time_zone
```
emerencia, line 147: idem

emerencia:
The rest LGTM

emerencia (approved):

projects/sport-data-valley/manual_seeds/rheuma_responses.rb
m-rios: that function adds .to_s

lib/tasks/manual_seed.rake
m-rios: it does

projects/sport-data-valley/manual_seeds/rheuma_responses.rb
m-rios: ok, i'll make it such that we don't get overlapped responses too

PR #2303, opened 2021-12-14
[43675857] Add rheumatism protocol and questionnaire

emerencia:
Fixes https://freedcamp.com/view/2354427/tasks/panel/task/43675857

Also fixes a weird rounding error that I found while generating ticks (below is the before picture):
ConEmu64_2021-12-14_18-26-56

### DIFF
--- a/projects/sport-data-valley/seeds/questionnaires/daily_questionnaire_rheumatism.rb
+++ b/projects/sport-data-valley/seeds/questionnaires/daily_questionnaire_rheumatism.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 db_title = ''
-db_name1 = 'Dagelijkse vragenlijst reuma'
+db_name1 = 'daily_questionnaire_rheumatism'
 dagboek1 = Questionnaire.find_by(key: File.basename(__FILE__)[0...-3])
 dagboek1 ||= Questionnaire.new(key: File.basename(__FILE__)[0...-3])
 dagboek1.name = db_name1
--- a/projects/sport-data-valley/seeds/questionnaires/daily_questionnaire_rheumatism.rb
+++ b/projects/sport-data-valley/seeds/questionnaires/daily_questionnaire_rheumatism.rb
@@ -1,0 +1,105 @@
+# frozen_string_literal: true
+
+db_title = ''
+db_name1 = 'Dagelijkse vragenlijst reuma'
+dagboek1 = Questionnaire.find_by(key: File.basename(__FILE__)[0...-3])
+dagboek1 ||= Questionnaire.new(key: File.basename(__FILE__)[0...-3])
+dagboek1.name = db_name1
+
+class RheumatismMethods
+  class << self
+    def category_title(category, title)
+      "<strong>#{category}</strong>: #{title}"
+    end
+  end
+end
+
+dagboek_content = [
+  {
+    id: :v1,
+    type: :range,
+    title: RheumatismMethods.category_title('Vermoeidheid', 'Hoe hevig was uw vermoeidheid gemiddeld vandaag?'),
+    min: 0,
+    max: 10,
+    step: 0.1,
+    required: true,
+    no_initial_thumb: true,
+    labels: ['geen vermoeidheid', 'meest voorstelbare vermoeidheid']
+  }, {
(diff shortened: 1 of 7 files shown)


projects/sport-data-valley/seeds/questionnaires/daily_questionnaire_rheumatism.rb
```diff
@@ -0,0 +1,105 @@
...
+    max: 10,
+    step: 0.1,
+    required: true,
+    no_initial_thumb: true,
+    labels: ['geen vermoeidheid', 'meest voorstelbare vermoeidheid']
+  }, {
+    id: :v2,
+    type: :radio,
+    title: RheumatismMethods.category_title('Vermoeidheid', 'Was er vandaag sprake van een ernstige piek in vermoeidheid?'),
+    options: ['Nee'],
```
m-rios, line 32: shouldn't Ja also be an option? or how does that work?

m-rios (approved):

projects/sport-data-valley/seeds/questionnaires/daily_questionnaire_rheumatism.rb
```diff
@@ -0,0 +1,105 @@
+# frozen_string_literal: true
+
+db_title = ''
+db_name1 = 'Dagelijkse vragenlijst reuma'
```
m-rios, line 4: all the other questionnaires have a snake_case name: e.g. `daily_questionnaire_zondag`

emerencia: We don't actually use the name anywhere. On base platform we use the key. The original idea was that the key has to be in filename format and could be symbolized, so only lowercase underscores and digits. Typically identical to the file name of the questionnaire to avoid confusion.

The name of a questionnaire was supposed to be a human readable description of a questionnaire, and only internally visible, e.g., to the researchers when previewing the questionnaire or when doing an export for this questionnaire.

Also I checked, and on SDV we use the questionnaire key to try and translate the name of hte questionnaire in the questionnaire manager interface.

But I do agree with your point that we should be consistent here so I'll change it.

emerencia: Fixed.

projects/sport-data-valley/seeds/questionnaires/daily_questionnaire_rheumatism.rb
emerencia: So "Ja" is the otherwise option here. Because when you select "Ja", you have to specify a reason in a text input. And they didn't describe that as being a separate question, just as a way to answer this question. The only way to do that in the questionnaire engine currently is by using the "otherwise" option, and then overwriting the default otherwise text with whatever you specify.

It looks like this:
chrome_2021-12-15_21-14-44

emerencia: And if you store it, it looks like this

```ruby
{"locale"=>"nl",
 "v1"=>"5.8",
 "v2"=>"Ja<br />Zo ja, waar denkt u dat dit door kwam?",
 "v2_ja_br_zo_ja_waar_denkt_u_dat_dit_door_kwam_text"=>"geen idee",
 "v3"=>"2.8",
 "v4"=>"7.4",
 "v5"=>"4",
 "v6"=>"2",
 "v7"=>"5.5",
 "v8"=>"7",
 "v1_timing"=>"37793",
 "v2_nee_timing"=>"1648",
 "v3_timing"=>"1132",
 "v4_timing"=>"1046",
 "v5_timing"=>"1558",
 "v6_timing"=>"1811",
 "v7_timing"=>"1257",
 "v8_timing"=>"3654",
 "v2_ja_br_zo_ja_waar_denkt_u_dat_dit_door_kwam_timing"=>"82",
 "v2_ja_br_zo_ja_waar_denkt_u_dat_dit_door_kwam_text_timing"=>"3961"}
```

emerencia: You can just check if the answer for v2 starts with "Ja"

PR #2306, opened 2021-12-17
[43610192] Add more rheuma questionnaires

emerencia:
Fixes https://freedcamp.com/view/2354427/tasks/panel/task/43610192


projects/sport-data-valley/seeds/protocols/rheumatism_one_time.rb
```diff
@@ -0,0 +1,54 @@
...
+bp_push_subscription = protocol.push_subscriptions.find_by(name: bp_name)
+bp_push_subscription ||= protocol.push_subscriptions.build(name: bp_name)
+bp_push_subscription.method = 'POST'
+bp_push_subscription.url = ENV['PUSH_SUBSCRIPTION_URL']
+bp_push_subscription.save!
+
+protocol.save!
+
+unused_measurement_ids = protocol.measurements.pluck(:id).to_set
+questionnaire_keys.each_with_index do |questionnaire_key, idx|
```
m-rios, line 24: based on the ticket, I understand that there should be a protocol per questionnaire. Are you sure this is how they want it?

m-rios (approved):

projects/sport-data-valley/seeds/protocols/rheumatism_one_time.rb
emerencia: I asked Dennis about this if he wanted one protocol per questionnaire or a single one for all of them. He said that for now, 1 protocol is good. Later they would want to pull them apart because a few are probably suited for a wider audience.

PR #2436, opened 2022-02-21
[44898748] fix html escape possible injection

emerencia:
Fixes https://freedcamp.com/view/2354427/tasks/panel/task/44898748


frbl:
Lgtm

PR #2442, opened 2022-02-28
[45179699] Fix seeds so demo user doesn't get more prot subs

emerencia:
Otherwise whenever we run the seeds, it creates a protocol subscription for the demo user that doesn't belong to any group.

Fixes https://freedcamp.com/view/2354427/tasks/panel/task/45179699


frbl:
lgtm

PR #2468, opened 2022-03-17
feat: add basic auth route to create a person

m-rios:
# Description of feature
...

# Checklist

### DIFF
--- a/app/controllers/api/v1/basic_auth_api/person_controller.rb
+++ b/app/controllers/api/v1/basic_auth_api/person_controller.rb
@@ -8,21 +8,23 @@ module Api
         before_action :set_mentor_role, only: %i[change_to_mentor]
 
         def create
-          auth_user = AuthUser.from_token_payload(new_person_params)
+          begin
+            auth_user = AuthUser.from_token_payload(new_person_params)
+          rescue RuntimeError => e
+            return validation_error(e.message) if team_error?(e)
+            # If we don't recognize the error raise it so AppSignal notifies us about it
+            raise e
+          end
           # This endpoint was added for UMO, where we need to sync users and we need their real emails stored so we can
           # invite them to fill in questionnaires. AuthUser.from_token_payload creates a new anonymous user (with no
           # email), so we need to add it manually
-          auth_user.person.update!(email: new_person_params[Rails.application.config.settings.metadata_field][:email])
+          auth_user.person.update(email: new_person_params[Rails.application.config.settings.metadata_field][:email])
 
-          render status: :created, json: auth_user
+          return created(auth_user.person) if auth_user.person.valid?
+
+          validation_error(auth_user.person.errors)
         rescue ActionController::ParameterMissing => e
-          # Do we have a standardized way of presenting errors?
-          render status: :bad_request, json: { error: e.message }.to_json
-        rescue StandardError => e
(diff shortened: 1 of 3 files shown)


app/controllers/api/v1/basic_auth_api/person_controller.rb
```diff
@@ -7,6 +7,24 @@ class PersonController < BasicAuthApiController
...
+        def create
+          auth_user = AuthUser.from_token_payload(new_person_params)
+          # This endpoint was added for UMO, where we need to sync users and we need their real emails stored so we can
+          # invite them to fill in questionnaires. AuthUser.from_token_payload creates a new anonymous user (with no
+          # email), so we need to add it manually
+          auth_user.person.update!(email: new_person_params[Rails.application.config.settings.metadata_field][:email])
+
+          render status: :created, json: auth_user
+        rescue ActionController::ParameterMissing => e
+          # Do we have a standardized way of presenting errors?
```
emerencia, line 19: We have the same `render_resource`, `validation_error`, `access_denied`, `not_found`, `created`, `destroyed`, `unprocessable_entity`, and `not_found` methods that we have in sdv base platform. They are defined in ApplicationHelper and included in the ApplicationController and so available in all controllers.

app/controllers/api/v1/basic_auth_api/person_controller.rb
```diff
@@ -7,6 +7,24 @@ class PersonController < BasicAuthApiController
...
+          render status: :bad_request, json: { error: e.message }.to_json
+        rescue StandardError => e
```
emerencia, line 21: Are they standard errors that you are capturing here? Or can we just capture a subclass. I don't like to capture all StandardErrors and then feed it back to the user because it might hide any actual problems that now won't be sent to AppSignal so we'll never know about them.

So at the very least, if you're sending back an internal server error, it should go to AppSignal as well.

But I rather have that we figure out exactly which errors there can be thrown here and capture only those. So like, those error messages that you're capturing, they're probably ActiveRecord::ValidationError (or something like that), right? Can we just capture only those?

Also, is there a way of calling `.update` on the person (instead of `.update!`) and then looking if the person is `.valid?` and returning its `.errors` a better way to do this? Or are all the errors thrown when creating the AuthUser? In that case can we put the try catch thing around just the AuthUser creation part and use `.update` instead of `.update!` on the person and handle at least those cases normally?

app/controllers/api/v1/basic_auth_api/person_controller.rb
m-rios: ah, that's what i was looking for. I remembered this were defined in application_controller.rb on sdv, but after a quick look on vsv i didn't find them. I'll try to them then.

app/controllers/api/v1/basic_auth_api/person_controller.rb
m-rios: the main reason why this is here is to catch exceptions from AuthUser yeah. It calls CreateAnonymousUser, which raises a bunch of errors when the metadata paremeters are invalid. The exception they raise are RuntimeExceptions. I can catch only those and move it inside the method.

emerencia:
LGTM

emerencia (approved):
LGTM

PR #2479, opened 2022-03-23
[43651009] Delete user from vsv via api

frbl:
# Description of feature
Delete users via API
...

# Checklist
- [NA] (If applicable) added example values of new ENV variables to .env.local and explained them in README.md.
- [X] Added unit tests to test the code.
- [X] Added integration tests to test the code within the application as a whole.

### DIFF
--- a/spec/integration/api/v1/basic_auth_api/person_controller_spec.rb
+++ b/spec/integration/api/v1/basic_auth_api/person_controller_spec.rb
@@ -32,21 +32,21 @@ describe 'Person API' do
     end
   end
   path '/basic_auth_api/person/{id}' do
-    destroy 'destroys a user' do
+    delete 'destroys a user' do
       tags 'Person'
       produces 'application/json'
       security [BasicAuth: {}]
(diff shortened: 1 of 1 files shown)


app/jobs/destroy_person_job.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+class DestroyPersonJob < ApplicationJob
+  queue_as :default
+
+  def perform(id)
+    person(id)&.destroy
```
emerencia, line 7: Shouldn't this be `.destroy!`? So that the job raises an error which is reported to appsignal and then the job will try again in case destroying fails?

emerencia, line 7: Shouldn't this be `.destroy!`? So that the job raises an error which is reported to appsignal and then the job will try again in case destroying fails?

```suggestion
    person(id)&.destroy
```

emerencia:
The rest LGTM

emerencia (approved):

PR #2483, opened 2022-03-28
Add note on having a CI background job

rikhuijzer:
I'm using CI for a year or so now. Works great for me. Easy to press the button in case I have to suddenly run the jobs.


frbl:
LGTM

PR #2494, opened 2022-04-04
[43635403] create protocols endpoint

m-rios:
# Description of feature
...

# Checklist


app/use_cases/create_or_update_protocol.rb
```diff
@@ -0,0 +1,97 @@
...
+        boolean :stop_measurement
+        boolean :should_invite
+        boolean :only_redirect_if_nothing_else_ready
+      end
+    end
+  end
+
+  validates :name, presence: true
+  validates :duration, presence: true
+  validates :invitation_text, presence: true
```
emerencia, line 30: Why are all these validations needed? Doesn't just having a property put that validation in there and also the error messages?

For example, if I look at the `CalculateScores` use case, which requires two arguments, a `questionnaire` hash an a `content` hash, then if I leave out the `questionnaire` hash, I get the following error when calling `.run!`:
ConEmu64_2022-04-06_17-45-27

and if I leave out exclamation point and just collect the .errors array, then I just get the error message for the attribute that I missed.

So why the `validates` statements?

app/use_cases/create_or_update_protocol.rb
```diff
@@ -0,0 +1,97 @@
...
+    @measurement.open_from_day = params[:open_from_day]
+    @measurement.period = params[:period]
+    @measurement.open_duration = params[:open_duration]
+    @measurement.reminder_delay = params[:reminder_delay]
+    @measurement.priority = params[:priority]
+    @measurement.stop_measurement = params[:stop_measurement]
+    @measurement.should_invite = params[:should_invite]
+    @measurement.only_redirect_if_nothing_else_ready = params[:only_redirect_if_nothing_else_ready]
+
+    # TODO: do not ! here either
```
emerencia, line 94: Can remove?

config/environments/development.rb
```diff
@@ -74,17 +74,15 @@
...
   # }
-  config.action_mailer.delivery_method = :mailgun
-  config.action_mailer.mailgun_settings = {
-    api_host: ENV['MAILGUN_API_HOST'],
-    api_key: ENV['MAILGUN_API_KEY'],
-    domain: ENV['MAILGUN_DOMAIN'],
-  }
+
+  config.action_mailer.delivery_method = :smtp
+  config.action_mailer.smtp_settings = { address: 'mailcatcher', port: 1025 }
```
emerencia, line 85: Could you add a mailcatcher to the docker-compose file, just so that this works out of the box for someone typing `docker-compose up`?

app/use_cases/create_or_update_protocol.rb
```diff
@@ -0,0 +1,97 @@
...
+  validates :name, presence: true
+  validates :duration, presence: true
+  validates :invitation_text, presence: true
+
+  validates :questionnaires, presence: true
+
+  def execute
+    return errors.merge!(@protocol.errors) unless initialize_protocol
+
+    return if set_measurements
```
emerencia, line 37: Shouldn't this return the @protocol ?

app/use_cases/create_or_update_protocol.rb
```diff
@@ -0,0 +1,97 @@
...
+      hash :measurement do
+        integer :open_from_offset
+        string :open_from_day, default: nil
+        integer :period, default: nil
+        integer :open_duration, default: nil
+        integer :reminder_delay, default: nil
+        integer :priority
+        boolean :stop_measurement
+        boolean :should_invite
+        boolean :only_redirect_if_nothing_else_ready
```
emerencia, line 23: Should we add `redirect_url` to this list? It's used to redirect the user back to wherever when they finished filling out all questionnaires open for them.

emerencia:
The rest LGTM. Specs are failing though.

emerencia (approved):

app/use_cases/create_or_update_protocol.rb
m-rios: you are right, the mere fact of declaring the properties indicates their presence is needed. TIL.

app/use_cases/create_or_update_protocol.rb
m-rios: done

config/environments/development.rb
m-rios: good idea

app/use_cases/create_or_update_protocol.rb
```diff
@@ -0,0 +1,97 @@
...
+
+  validates :questionnaires, presence: true
+
+  def execute
+    return errors.merge!(@protocol.errors) unless initialize_protocol
+
+    return if set_measurements
+
+    errors.merge!(@measurement.errors) if @measurement.present?
+  end
```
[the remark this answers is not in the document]
m-rios: YES! Great catch, this is why I wanted you to review it. I've done exactly what you've suggested, and wrote a spec

app/controllers/api/v1/basic_auth_api/protocols_controller.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    module BasicAuthApi
+      class ProtocolsController < BasicAuthApiController
+        def create
+          res = CreateOrUpdateProtocol.run **update_protocol_params
+          return created(res.protocol) if res.valid?
```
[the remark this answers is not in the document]
m-rios: active interactions don't return what the run function returns, they return an active interaction job. Think of it as a model (it is very similar). From the object you can get access to stuff like errors (so you can do full error management without raising, if you want to), and since it's a PORO, you can make attributes accessible (see line 9 in CreateOrUpdateProtocol)

app/use_cases/create_or_update_protocol.rb
m-rios: see comment above

app/use_cases/create_or_update_protocol.rb
m-rios: Yeah, probably, I didn't add it because I wasn't using it, but for other projects we might need it

app/controllers/api/v1/basic_auth_api/protocols_controller.rb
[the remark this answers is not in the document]
emerencia: Ah, now that you mention it, I remember that is how it was.

PR #2497, opened 2022-04-06
[39147365] prefill questionnaires 

m-rios:
# Description of feature
...

# Checklist


app/generators/dropdown_generator.rb
```diff
@@ -29,17 +29,23 @@ def generate_dropdown(question, id)
...
     option_options = {
-      value: option[:raw][:value].presence || option[:raw][:title]
+      value: value
     }
     option_options = add_shows_hides_questions(option_options, option[:shows_questions], option[:hides_questions])
+    option_options = decorate_with_previous_value(
+      question,
+      idify(question[:id]),
+      option_options
+    ) { |previous_value| [:selected, previous_value == value] }
```
emerencia, line 48: very nice

emerencia:
So in relation to this, Frank's code specifically allows you to replace question text with text from a previous response, explicitly not to prefill answers. So there is  a distinction.

app/generators/concerns/prefillable.rb
```diff
@@ -0,0 +1,30 @@
...
+            .first
+  end
+
+  def previous_response_content(response)
+    prev_response = previous_response(response)
+    return nil unless prev_response.present?
+
+    prev_response.remote_content&.content
+  end
+end
```
emerencia, line 30: So in relation to this, Frank some time ago added code to use a previous response's answers to use in the text of a questionnaire question, i.e., it would be used in a context saying "Previously, you answered the following question with the text: {{previous_v14}}." The questions would be annotated with a `uses: ` clause that loaded a previous response's values up for use in the question texts.

Now this differs from your case, because in Frank's case, we're replacing text in a question with text from a response, whereas in your case, we're prefilling a questionnaire with values from a previous response.

My only concern is that the code in `app/logic/previous_response_finder.rb` shares a lot of logic with `app/generators/concerns/prefillable.rb`. Essentially it's code duplication, although I believe your version to be better (with the nice SQL query and such). Is there a way you can change `app/logic/previous_reponse_finder.rb` to use your prefillable concern instead of duplicating the same logic?

app/generators/question_type_generator.rb
```diff
@@ -146,20 +146,43 @@ def add_raw_to_option(option, question, idx)
...
     tag.div(option_field, class: 'input-field inline')
   end
 
+  def otherwise_textfield_options(question)
+    id = idify(question[:id], question[:raw][:otherwise_label], 'text')
+    decorate_with_previous_value(
+      question,
+      id, {
+        id: id,
+        name: answer_name(idify(question[:id], question[:raw][:otherwise_label], 'text')),
```
emerencia, line 161:
```suggestion
        name: answer_name(id),
```

app/generators/radio_generator.rb
```diff
@@ -49,14 +49,19 @@ def decorate_with_stop_subscription(question, option_body, option)
...
+      question[:id],
+      {
+        name: answer_name(idify(question[:id])),
+        type: 'radio',
+        id: elem_id,
+        value: value,
+        required: !(question.key?(:required) && question[:required] == false),
+        class: 'validate'
+      }
+    ) { |previous_value| [:checked, previous_value == value] }
```
emerencia, line 64: A good test to add (if you don't have it already) might be to have a radio question with a prefilled option selected that has a `shows_questions` attribute, and then checking if the questions in that `shows_questions` thing actually show up when the questionnaire is loaded (I think they should, but it is something we'd want to test).

app/generators/radio_generator.rb
```diff
@@ -67,11 +72,18 @@ def radio_otherwise(question)
...
+        { name: answer_name(idify(question[:id])),
+          type: 'radio',
+          id: idify(question[:id], question[:raw][:otherwise_label]),
+          value: value,
+          required: true,
+          class: 'otherwise-option' }
+      ) { |previous_value| [:checked, previous_value == value] }
+    )
   end
 end
```
emerencia, line 89: Why not add support for the checkbox question type also? I understand leaving out the support for sliders, because sliders are quickly and easy to set by the user, but a checkbox seems to me no different than a dropdown or radio question?

app/generators/checkbox_generator.rb
```diff
@@ -28,18 +28,20 @@ def answer_options(question)
...
+    previous_value = question[:previous_response]&.[](elem_id)
     {
       type: 'checkbox',
       id: elem_id,
       name: answer_name(elem_id),
-      value: true
+      value: true,
+      checked: previous_value.presence || nil
     }
   end
```
emerencia, line 40: What about the otherwise option for checkboxes? I don't think it's being prefilled if someone selects "otherwise" for a checkbox?

app/generators/checkbox_generator.rb
```diff
@@ -28,18 +28,20 @@ def answer_options(question)
...
-  def question_options(elem_id)
+  def question_options(question, elem_id)
+    previous_value = question[:previous_response]&.[](elem_id)
     {
       type: 'checkbox',
       id: elem_id,
       name: answer_name(elem_id),
-      value: true
+      value: true,
+      checked: previous_value.presence || nil
```
emerencia, line 38: Couldn't you also use the `decorate_with_previous_value` method here? with a yield and then you return like the `[:checked, previous_value.presence || nil]` here? Passing `elem_id` instead of `id`? It would make the code a bit more consistent and then the `question[:previous_response]&.[](...)` logic of how to retrieve the previous value stays isolated in one place (the `decorate_with_previous_value` method).

emerencia (approved):
The rest LGTM. Very nice addition I think.

app/generators/checkbox_generator.rb
m-rios: oh yes, I implemented the decorator after this but forgot to replace it

app/generators/checkbox_generator.rb
m-rios: yes, good point, missed it

app/generators/concerns/prefillable.rb
m-rios: wow, the overlap is uncanny. I didn't know something like this existed. I'll try to merge it somehow

m-rios: k, came up with something

PR #2512, opened 2022-04-13
[45812742] Language Questionnaire

m-rios:
# Description of feature
...

# Checklist


app/controllers/questionnaire_controller.rb
```diff
@@ -75,6 +76,13 @@ def create_informed_consent
     redirect_to questionnaire_path(uuid: @response.uuid)
   end
 
+  def set_language_preference
+    selected_locale = questionnaire_content['v0']
+    current_user.update! locale: selected_locale == 'Nederlands' ? 'nl' : 'en'
+    @response.invitation_set.update! locale_retrieved: true
+    redirect_to NextPageFinder.get_next_page current_user: current_user
+  end
+
```
emerencia, line 85: Wait, so it sets the user locale based on something that is encoded in the questionnaire, and then afterwards it doesn't set it back? I'm not sure if it is the best solution. Couldn't we do something in the `questionnaire_generator.rb` itself that when it sees there's a "v0" key (which I would prefer if we just had maybe a "lang" key in the questionnaires, and leave the `v` prefix for actual questions), but if it had such a key that it then would override its given `locale`  param (see line 13 in `app/generators/questionnaire_generator.rb`). Would that work for this intended purpose?

app/models/protocol.rb
```diff
@@ -7,6 +7,7 @@ class Protocol < ApplicationRecord
   validate :at_most_one_stop_measurement
   has_many :protocol_subscriptions, dependent: :destroy
   belongs_to :informed_consent_questionnaire, class_name: 'Questionnaire', optional: true # can be nil
+  belongs_to :language_questionnaire, class_name: 'Questionnaire', optional: true # can be nil
```
emerencia, line 10: Instead of making an entire language questionnaire. Would it suffice to simply just make a route where the user can set their locale (as in not using the questionnaire rendering engine, but just a simple page to update one property of their `protocol_subscription`, prefilled with the default setting of their `person`)? What I mean is some setting in a `protocol_subscription` that is a simple boolean that checks, hey did the user already specify a locale that they'd like to use for this protocol, if yes, then fine, if no, redirect them to /language or whatever, which is a site where they can choose between English or Dutch (since those are only two locales supported by the platform), and can be the same for all questionnaires and protocols I think, and then upon saving, do something with that callback that saves the setting for their protocol subscription and then moves on?

Because the proposed approach feels too much like a "double informed consent" to me. If we need to ask anything more than a simple language question -> use an informed questionnaire, and otherwise the page where you set a language can be the same across all protocols, I think.

m-rios: mm, i see what you're saying. I like it.

app/controllers/languages_controller.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+class LanguagesController < ApplicationController
+  # protect_from_forgery prepend: true, with: :exception, except: :create
+  # skip_before_action :verify_authenticity_token, only: %i[interactive_render from_json]
+  # before_action :log_csrf_error, only: %i[create]
+  # before_action :set_response, only: %i[show preference destroy]
+  # before_action :set_locale, only: %i[show]
```
emerencia, line 8: Can remove these lines?

db/migrate/20220519094533_add_language_preferences_to_protocol_subscription.rb
```diff
@@ -0,0 +1,6 @@
+class AddLanguagePreferencesToProtocolSubscription < ActiveRecord::Migration[6.1]
+  def change
+    add_column :protocol_subscriptions, :needs_language_input, :boolean, null: false, default: false
```
emerencia, line 3: The thing that seems weird to me is why not define the `needs_language_input` on the `protocol` level, rather than on the `protocol_subscription` level? I agree that `has_language_input` should be on the `protocol_subscription` level, but surely you'd just want say "all subscriptions for this protocol should first pick their language", rather than setting it for each new protocol subscription?

What am I missing here?

app/controllers/languages_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+      Rails.logger.error "invalid user: #{current_user.errors.full_messages}"
+      return flash.alert = I18n.t('pages.languages.flash_messages.alert.locale_update_error')
+    end
+
+    unless @response.protocol_subscription.update has_language_input: true
+      Rails.logger.error "invalid protocol_subscription: #{@response.errors.full_messages}"
+      return flash.alert = I18n.t('pages.languages.flash_messages.alert.locale_update_error')
+    end
+
+    redirect_to change_params[:cb], notice: I18n.t('pages.languages.flash_messages.notice.locale_updated')
```
emerencia, line 30: can `change_params[:cb]` be nil? and if so, does `redirect_to nil` redirect to the root? Or does it not redirect? or does it give an error?

app/controllers/questionnaire_controller.rb
```diff
@@ -411,4 +412,10 @@ def set_default_content
     # Check if the parsing was wrong, if it was, we don't do anything. If it was something else, reraise.
     raise e if e.message != 'invalid base64'
   end
+
+  def check_language
+    subscription = @response.protocol_subscription
+    language_must_be_set = subscription.needs_language_input && !subscription.has_language_input
```
emerencia, line 418: The needs_language_input really feels like a protocol setting to me.

app/views/languages/show.html.haml
```diff
@@ -0,0 +1,15 @@
...
+    .row
+      .col.s12
+        - locales.each do |locale|
+          %p
+            %label
+              %input{name: 'locale', type: 'radio', value: locale, checked: @locale == locale}
+              %span=locale
+    .row
+      .col.s12
+        = button_tag('Submit', type: 'submit', class: 'btn waves-effect waves-light')
```
emerencia, line 15: Can we i18nnify this "Submit" prompt? I think there's probably a translation already but the word in Dutch would be "Opslaan"

config/locales/pages/languages/nl.yml
```diff
@@ -0,0 +1,8 @@
+nl:
+  pages:
+    languages:
+      flash_messages:
+        alert:
+          response_not_found: Response not found
```
emerencia, line 6:
```suggestion
          response_not_found: Antwoord niet gevonden
```

config/locales/pages/languages/nl.yml
```diff
@@ -0,0 +1,8 @@
+nl:
+  pages:
+    languages:
+      flash_messages:
+        alert:
```
emerencia, line 5: It's missing a translation for `locale_udpate_error`

config/locales/pages/languages/en.yml
```diff
@@ -0,0 +1,9 @@
+en:
+  pages:
+    languages:
+      flash_messages:
+        alert:
+          response_not_found: Response not found
+          locale_udpate_error: Unable to update the locale preferences
```
emerencia, line 7:
```suggestion
          locale_update_error: Unable to update the locale preferences
```
?

config/locales/pages/languages/nl.yml
```diff
@@ -0,0 +1,8 @@
...
+          response_not_found: Response not found
+        notice:
+          locale_updated: Language preference has been changed
```
emerencia, line 8:
```suggestion
          locale_updated: Taalinstelling opgeslagen
```

emerencia:
The rest LGTM!

app/controllers/languages_controller.rb
m-rios: not for my use case, but if this gets reused then potentially, yes. I'll think about what to do in that case, good catch

m-rios: it fails if it's nil. I've changed it to only redirect if a cb is given

app/views/languages/show.html.haml
m-rios: Oops, i missed it, thanks

PR #2529, opened 2022-04-29
Fix silencer bump to 2.0

rikhuijzer:
Note that this PR is pointed against #2524.

This PR is basically just a copy of the [silencer README](https://github.com/stve/silencer).


frbl:
Lgtm

PR #2519, opened 2022-04-21
feat: add elastic apm agent

m-rios:
# Description of feature
...

# Checklist
- [ ] (If applicable) added example values of new ENV variables to .env.local and explained them in README.md.
- [ ] Added unit tests to test the code.
- [ ] Added integration tests to test the code within the application as a whole.

PR #2537, opened 2022-05-08
fix: error message when elastic apm env vars are not set

m-rios:
# Description of feature
The elastic APM agent will now be disabled by default.

# Checklist

### DIFF
--- a/config/elastic_apm.yml
+++ b/config/elastic_apm.yml
@@ -1,3 +1,4 @@
+enabled: <%= ENV["ELASTIC_APM_ENABLED"] == 'true'%>
 server_url: <%= ENV["ELASTIC_APM_URL"] %>
 secret_token: <%= ENV["ELASTIC_APM_SECRET_TOKEN"] %>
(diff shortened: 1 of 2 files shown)


m-rios:
@rikhuijzer This should fix the issue you raised here https://github.com/compsy/u-can-act/pull/2519

config/elastic_apm.yml
```diff
@@ -1,3 +1,4 @@
+enabled: <%= ENV["ELASTIC_APM_ENABLED"] == 'true'%>
```
rikhuijzer, line 1: Yes. Nice. According to the elastic java docs (first one I could find):

> Setting to false will completely disable the agent, including instrumentation and remote config polling. If you want to dynamically change the status of the agent, use [recording](https://www.elastic.co/guide/en/apm/agent/java/master/config-core.html#config-recording) instead.

Have you also tested whether the system can now run without having the other environment variables set? If yes, then I approve this PR.

m-rios: I ran it locally with RAILS_ENV set to production and got no errors.

PR #2804, opened 2022-11-09
[49098472] Psychological competencies

emerencia:


projects/sport-data-valley/seeds/protocols/psychological_competencies.rb
```diff
@@ -0,0 +1,36 @@
+default_protocol_duration = 1.day  # evt eerder dynamisch afbreken
```
m-rios, line 1: just 1 day, so it's a one-off questionnaire?

projects/sport-data-valley/seeds/protocols/psychological_competencies.rb
```diff
@@ -0,0 +1,36 @@
...
+
+pr_name = 'psychological_competencies'
+protocol = Protocol.find_by(name: pr_name)
+protocol ||= Protocol.new(name: pr_name)
+
+protocol.duration = default_protocol_duration
+protocol.informed_consent_questionnaire = nil
+protocol.invitation_text = 'Je bent uitgenodigd door je coach om een vragenlijst in te vullen.'
```
m-rios, line 9: do we need to translate this? how do we do it on the other questionnaires?

projects/sport-data-valley/seeds/questionnaires/psychological_competencies.rb
```diff
@@ -0,0 +1,416 @@
...
+db_name1 = 'psychological_competencies'
+dagboek1 = Questionnaire.find_by(key: File.basename(__FILE__)[0...-3])
+dagboek1 ||= Questionnaire.new(key: File.basename(__FILE__)[0...-3])
+dagboek1.name = db_name1
+
+class PsychologicalCompetenciesMethods
+  DEFAULT_QUESTION_OPTIONS = {
+    type: :likert,
+    options: [
+      { title: 'Helemaal mee oneens', numeric_value: 0 },
```
m-rios, line 13: I'm only seeing dutch text everywhere. I'm beginning to think that this questionnaire does not need translations. Correct?

projects/sport-data-valley/seeds/questionnaires/psychological_competencies.rb
```diff
@@ -0,0 +1,416 @@
...
+      en: ""
+    }
+  }, {
+    section_start: 'Psychologische competentievragenlijst',
+    id: :v1,
+    type: :radio,
+    title: 'Ik verklaar op een voor mij duidelijke wijze te zijn ingelicht over aard, methode, doel, risico’s en de belasting van het onderzoek. Ik ga ermee akkoord dat de verzamelde gegevens en resultaten gebruikt kunnen worden voor wetenschappelijk onderzoek.',
```
m-rios, line 57: so you disabled the informed consent at the protocol level, and made it part of the questionnaire itself. Is it because they needed something custom?

m-rios (approved):

projects/sport-data-valley/seeds/protocols/psychological_competencies.rb
emerencia: Yes, it is a one-off questionnaire.

projects/sport-data-valley/seeds/protocols/psychological_competencies.rb
emerencia: Good question. Let me look this up.

emerencia: Okay, so this is just the default invitation text if none is provided. However, it can be overridden when starting a protocol from base platform (you then specify both an English and a Dutch invitation text).

emerencia: I'll add a comment.

emerencia: Fixed.

projects/sport-data-valley/seeds/questionnaires/psychological_competencies.rb
emerencia: It does, but that will come in a second iteration. I talked to Dennis and what we need live for Wednesday is just the Dutch version. I'll do the English version ASAP afterwards.

projects/sport-data-valley/seeds/questionnaires/psychological_competencies.rb
emerencia: Sot this is a straight import of a copy of a questionnaire that they had in a different system, where they solved this in the way that it is now. I think ideally we would split up the informed consent and questionnaire into separate things, but the way it's phrased here, with people being unable to continue if they didn't check all three boxes I think is fine.

PR #2828, opened 2022-11-26
Add second training log protocol with invites and without otr

emerencia:
Merge after: https://gitlab.com/researchable/sport-data-valley/mvp/base-platform/-/merge_requests/1882


projects/sport-data-valley/seeds/protocols/training_log_with_invites.rb
```diff
@@ -0,0 +1,34 @@
...
+protocol.save!
+
+bp_name = 'base-platform-subscription-after-training-with-invites'
+bp_push_subscription = protocol.push_subscriptions.find_by(name: bp_name)
+bp_push_subscription ||= protocol.push_subscriptions.build(name: bp_name)
+bp_push_subscription.method = 'POST'
+bp_push_subscription.url = ENV['PUSH_SUBSCRIPTION_URL']
+bp_push_subscription.save!
+
+protocol.save!
```
frbl, line 18: do you need 2 saves?

projects/sport-data-valley/seeds/protocols/training_log_with_invites.rb
```diff
@@ -0,0 +1,34 @@
...
+
+name = 'training_log'
+dagboekvragenlijst_id = Questionnaire.find_by(name: name)&.id
+raise "Cannot find questionnaire: #{name}" unless dagboekvragenlijst_id
```
frbl, line 22: bit nitpicky, but should this whole thing be in a transaction? If the chances of this happening are super slim, I wouldnt' worry about it too much

projects/sport-data-valley/seeds/protocols/training_log_with_invites.rb
```diff
@@ -0,0 +1,34 @@
...
+
+db_measurement = protocol.measurements.where(questionnaire_id: dagboekvragenlijst_id).first
+db_measurement ||= protocol.measurements.build(questionnaire_id: dagboekvragenlijst_id)
+db_measurement.period = nil
+db_measurement.open_duration = nil
+db_measurement.open_from_offset = 0
+db_measurement.reward_points = 0
+db_measurement.stop_measurement = true
```
frbl, line 30: this should indeed be true? (double check)

frbl (approved):
LGTM

projects/sport-data-valley/seeds/protocols/training_log_with_invites.rb
emerencia: Yes, because we start a new protocol subscription for every training log, and we want to end it exactly when the user fills out the training log.

projects/sport-data-valley/seeds/protocols/training_log_with_invites.rb
emerencia: Nope. Removed.

projects/sport-data-valley/seeds/protocols/training_log_with_invites.rb
emerencia: I don't think putting it in a transaction adds much because the entire thing is idempotent. So it doesn't matter where or if we fail, we are always in a good state.

PR #2927, opened 2023-03-08
Refactor/use helm chart for deploy

sjoukedv:
# Description of feature
Similar to https://gitlab.com/researchable/aukati/svc-auth we add a helm chart for better reusability (and easier deployment) with ArgoCD


emerencia:
We now have done this in a different branch.

PR #3420, opened 2024-06-12
feat: adds interactive / example questionnaire to support post requests

JPAntonisse:
# Description of feature

adds post url: /questionnaire/interactive_post
that is the same as /questionnaire/interactive, only accepts PSOT instead of GET


emerencia:
LGTM

PR #3529, opened 2024-10-09
[60921771] Add helm chart

emerencia:
Fixes https://freedcamp.com/view/3448368/tasks/panel/task/60921771


chart/develop/templates/cron.yaml
```diff
@@ -0,0 +1,306 @@
...
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          restartPolicy: Never
+          imagePullSecrets:
+            - name: {{ .Values.gitlabAuthSecretName }}
+          containers:
+          - name: cron-container
+            image: registry.gitlab.com/researchable/general/docker-images/curl:latest
```
maxverbeek, line 20: This is a private image, are you sure this shouldn't be configurable?

chart/develop/templates/deployment.yaml
```diff
@@ -0,0 +1,299 @@
...
+        track: stable
+        tier: web
+    spec:
+      imagePullSecrets:
+        - name: {{ .Values.gitlabAuthSecretName }}
+        - name: {{ .Values.dockerAuthSecretName }}
+      containers:
+        - name: svc-questionnaires
+          imagePullPolicy: Always
+          image: docker.io/researchableuser/sdv-svc-questionnaires:{{ .Values.tag }}
```
maxverbeek, line 32: This should be configurable from the yaml.

chart/develop/templates/cron.yaml
```diff
@@ -0,0 +1,306 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: svc-questionnaire-scheduler-daily-at-one-am
+  namespace: {{ .Values.namespace }}
```
maxverbeek, line 5: All of the resources should not have a namespace. Helm adds the namespace for you with the --namespace flag (which does not come from values.yaml). Idem with ArgoCD -- it gets the namespace from the destination, instead of values.yaml.

maxverbeek: Sorry if I told you earleir that you could put it in values.yaml :sweat_smile:

chart/develop/templates/deployment.yaml
```diff
@@ -0,0 +1,299 @@
...
+          - name: NODE_ENV
+            value: production
+          - name: REVISION
+            value: {{ .Values.commitSha }}
+          - name: APPSIGNAL_APP_ENV
+            value: {{ .Values.appsignalAppEnv }}
+          - name: APPSIGNAL_ACTIVE
+            value: "true"
+          - name: PAPERTRAIL_PROGRAM
+            value: questionnaires
```
maxverbeek, line 52: Since we want to get rid of papertrail, we can optionally wrap this in an if-statement so we can enable/disable this stuff from values.yaml. Optional though, we could always add this later.

chart/develop/templates/deployment.yaml
```diff
@@ -0,0 +1,299 @@
...
+          - name: PUSH_SUBSCRIPTION_URL
+            valueFrom:
+              secretKeyRef:
+                name: {{ .Values.secretName }}
+                key: PUSH_SUBSCRIPTION_URL
+          - name: BASE_PLATFORM_URL
+            valueFrom:
+              secretKeyRef:
+                name: {{ .Values.secretName }}
+                key: BASE_PLATFORM_URL
```
maxverbeek, line 67: There are a lot of secrets in here that should maybe be a configmap instead

chart/develop/templates/deployment_hpa.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: autoscaling/v2
+kind: HorizontalPodAutoscaler
```
maxverbeek, line 2: In addition to an HPA, can you also add a PodDisruptionBudget? (both deployment and the worker)

chart/develop/values.yaml
```diff
@@ -0,0 +1,7 @@
+secretName: svc-questionnaires-secret
+gitlabAuthSecretName: gitlab-auth
+dockerAuthSecretName: docker-auth
+tag: latest
+namespace: usc-chan
+commitSha: some-commit-sha
+appsignalAppEnv: staging
```
maxverbeek, line 7: We have this in svc-auth as well. Do you know why? Because I don't.. You can target separate versions of the helm chart using branches/refnames/commit tags already.

maxverbeek: In fact we do use refnames already, so this adds a lot of confusion because you can configure it to master-develop or develop-master.

maxverbeek (approved):
I just realised at the end that you have the same chart twice: once in develop and once in master. So my comments could be a bit spread out over both of these. My last comment is that we should just have a single chart like every other helm project does, and configure differences between develop/master via the normal way through ArgoCD.

otherwise lgtm

chart/develop/templates/cron.yaml
emerencia: Yes I'm sure. That's what we have the gitlabAuth for.

emerencia: Fine, I'll change it.

emerencia: Fixed.

chart/develop/templates/cron.yaml
emerencia: Ok so I just remove the namespace from everywhere?

emerencia: Fixed

chart/develop/templates/deployment.yaml
emerencia: I'll remove it altogether

emerencia: Fixed.

chart/develop/templates/deployment.yaml
emerencia: Dammit you're right. Ok I'll fix it.

emerencia: Fixed.

chart/develop/templates/deployment_hpa.yaml
emerencia: yes you're right althoguht i have no clue how to do that

emerencia: Ok Fixed.

chart/templates/deployment-budget.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
+  name: {{ .Release.Name }}-budget
+spec:
+  minAvailable: 1
+  selector:
+    matchLabels:
+      app: {{ .Release.Name}}
+      tier: web
```
maxverbeek, line 10: These labels are not correct for all resources. The Deployment resource has

```
app: svc-questionnaires
```

chart/templates/worker-budget.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
+  name: {{ .Release.Name }}-worker-budget
+spec:
+  minAvailable: 1
+  selector:
+    matchLabels:
+      app: {{ .Release.Name}}
+      tier: worker
```
maxverbeek, line 10: Same problem as the web PDB: it has app: svc-questionnaires hardcoded, maybe change that there

chart/values.yaml
```diff
@@ -0,0 +1,22 @@
+secretName: svc-questionnaires-secret
+gitlabAuthSecretName: gitlab-auth
+dockerAuthSecretName: docker-auth
```
maxverbeek, line 3: I have 1 last comment about this:

currently we are required to provide two secrets for image pull credentials: gitlab-auth and docker-auth. This is actually a hack because we were migrating from gitlab to docker. Now, when the migration is done we will forever be required to provide gitlab secrets because of this.

I suggest you create a list in values.yaml with imagePullSecretNames, and iterate over these secrets wherever they are used. That way, when we have migrated (which we already have) we don't need two secrets anymore.

chart/templates/worker.yaml
```diff
@@ -34,36 +33,27 @@ spec:
             requests:
```
maxverbeek, line 33: I cannot place a comment on the relevant line, but above this comment you still have

```
image: docker.io/researchableuser/sdv-svc-questionnaires:{{ .Values.tag }}
```

hardcoded to sdv

chart/templates/deployment.yaml
```diff
@@ -35,6 +34,9 @@ spec:
               memory: "1.5Gi"
```
maxverbeek, line 34: Same above this line: you have an image: ... sdv hardcoded still.

chart/templates/deployment.yaml
```diff
@@ -35,6 +34,9 @@ spec:
...
             requests:
               memory: "768Mi"
```
maxverbeek, line 36: I'm not sure if the resource limits & requests should be made configurable through Helm.. I guess the chart authors (compsy) can decide what values are sensible, but maybe these values have to scale according to expected workload of the application.

maxverbeek (approved):
nice Ando! I think after these changes the helm chart is looking pretty solid. We should aim for portability so that we can also move projects like SDV over to this helmchart.

chart/templates/deployment.yaml
maxverbeek: In either case I don't think it hurts to create values.yaml entries with these values as default. That way endusers of the helmchart can override them as they see fit.

chart/develop/values.yaml
emerencia: No I think this is the environment for appsignal. So when you have a project on app signal it marks it as either staging or production. To differentiate between the two. It can also use the branch name, but I don't know if this is available here?

chart/templates/deployment-budget.yaml
emerencia: Ok I'll hardcode it in there.

emerencia: Fixed.

chart/templates/worker-budget.yaml
emerencia: Fixed.

chart/templates/deployment-budget.yaml
emerencia: Edit; nvm: i changed the hardcoded one to use .Release.Name

chart/templates/worker.yaml
emerencia: Yes, that's because that's where the image is pushed to in Dockerhub in the Gitlab mirror of the GitHub repo. Strictly not correct, but we can leave it for now I think.

chart/templates/deployment.yaml
emerencia: Ok, I'll fix it.

emerencia: Fixed.

chart/templates/deployment.yaml
emerencia: Yeah I think ti's okay. THat's just how the image is named.

PR #3544, opened 2024-10-21
feat: split redis and mongo in separate secrets

maxverbeek:
# Description of feature
...

# Checklist
- [x] (If applicable) added example values of new ENV variables to .env.local and explained them in README.md.

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -1,7 +1,7 @@
 secretName: svc-questionnaires-secret
-databaseSecretName: svc-questionnaires-database-credentials
-redisSecretName: svc-questionnaires-redis-credentials
-mongoSecretName: svc-questionnaires-mongo-credentials
+# databaseSecretName: svc-questionnaires-database-credentials
+# redisSecretName: svc-questionnaires-redis-credentials
+# mongoSecretName: svc-questionnaires-mongo-credentials
 tag: latest
 appsignalAppEnv: staging
 curlImage: registry.gitlab.com/researchable/general/docker-images/curl:latest
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -1,5 +1,7 @@
 secretName: svc-questionnaires-secret
 databaseSecretName: svc-questionnaires-database-credentials
+redisSecretName: svc-questionnaires-redis-credentials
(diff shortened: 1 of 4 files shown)


chart/values.yaml
```diff
@@ -1,5 +1,7 @@
 secretName: svc-questionnaires-secret
 databaseSecretName: svc-questionnaires-database-credentials
+redisSecretName: svc-questionnaires-redis-credentials
+mongoSecretName: svc-questionnaires-mongo-credentials
```
emerencia, line 4: The way I would do this with my non-experience in writing Helm charts is just to use `{{ .Values.mongoSecretName }}` and `{{ .Values.redisSecretName }}` everywhere (instead of `{{ .Values.mongoSecretName | default .Values.secretName }}` and `{{ .Values.redisSecretName | default .Values.secretName }}` respectively), and then specify these values  like this:

```
secretName: svc-questionnaires-secret
databaseSecretName: svc-questionnaires-database-credentials
redisSecretName: svc-questionnaires-secret
mongoSecretName: svc-questionnaires-secret
```

So that it's clear that they're by default being set to the value of `secretName` (but you can still overwrite it), rather than you having to read the README or derive it from looking at the code.

But I'm curious to hear as to why you chose this approach instead.

emerencia (approved):
The rest LGTM.

chart/values.yaml
maxverbeek: Good point. I was thinking we should not make it mandatory for people to provide values for redisSecretName and mongoSecretName. I didn't want to assume everyone is using the default value that we currently set for secretName. e.g. if you change secretName in an override, you would break the configuration by assuming that the other secrets have also changed to the value that you set. (they still default to the value that was left in the default values.yaml).

maxverbeek: Ok now the secrets will default to the value you set for secretName, so if you set secretName to

```yaml
secretName: super-secret
```

then the databaseSecretName, etc will also be looked up from `super-secret` unless you explicitly override it by also specifying `databaseSecretName: super-secret-database`, and likewise for redis and mongo.

emerencia: Ok, nice.

PR #3545, opened 2024-10-21
feat: make connection string keys configurable

maxverbeek:
The motivation behind this is that the bitnami/redis chart creates a secret with a connection string, but the key in this secret is not defined the way the questionnaire engine expects it.

# Description of feature
...

# Checklist


emerencia (approved):
lgtm

PR #3609, opened 2024-12-16
add support to generate jwt tokens encrypted in rs256

pmkihara:
# Description of feature
- Add support to generate JWT tokens encrypted in RS256. This is needed to push a response to the base backend and ensure that the AuthUser is correctly identified there.
- Add a PushSubscription to USC Chan project
- Small CSS fix to avoid having the navy background in the edit person page (the user is redirected to this page after filling a questionnaire and they don't have a mobile phone or an email address set up yet)

# Checklist
- [x] (If applicable) added example values of new ENV variables to .env and explained them in README.md.


emerencia:
The rest LGTM, assuming you tested the RS256 stuff locally that it works with the JWTokens and stuff.

emerencia (approved):

pmkihara:
Yes, the RS256 stuff was tested locally and it integrates with the user from svc-auth and the profile from base-backend. The base-backend was able to identify the current_profile correctly with this setup.

pmkihara:
> Btw in the protocol, you can set a redirect URL, for the user to be redirected to after filling out a questionnaire. You can see some of the Sport Data Valley protocols for examples. So that way, the user never ends up in the edit person page. That really is just a default page for if there is no redirect URL, but I think here it's good to maybe redirect to a URL on the vitamaps app that says something like "Thanks for filling out the questionnaire" or whatever or maybe even redirects them directly to their results page with just a flash notification that says "Thanks for...".

I've updated the redirect url to redirect oto the vitamaps app with a query param to identify the origin of the redirect.

PR #3639, opened 2025-01-30
specs: change to new headless mode and adapt spec

pmkihara:
# Description of feature
Update the selenium options to use the new headless mode, since the old version doesn't seem to be supported any longer (it would give a `Selenium::WebDriver::Error::SessionNotCreatedError:`). 

# Checklist


emerencia (approved):
LGTM

PR #3653, opened 2025-02-10
feat: try to fix auth errors on umo

maxverbeek:
We're getting this error when previewing questionnaires from QMI:

image


maxverbeek:
Can you help us Ando? We're not sure if this is correct

emerencia (approved):
LGTM

PR #3674, opened 2025-03-06
Feature/mp standardize k8s labels

millenevprado:
# Description of feature
...
https://freedcamp.com/view/3272505/tasks/61596117
# Checklist

### DIFF
--- a/chart/templates/cron.yaml
+++ b/chart/templates/cron.yaml
@@ -3,8 +3,8 @@ kind: CronJob
 metadata:
   name: {{ .Release.Name }}-scheduler-daily-at-one-am
   labels:
-    app: svc-questionnaires-scheduler
-    observability.researchable.nl/job: {{ .Release.Name }}-scheduler-daily-at-one-am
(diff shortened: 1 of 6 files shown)


chart/templates/cron.yaml
```diff
@@ -2,16 +2,25 @@ apiVersion: batch/v1
...
 spec:
   schedule: "0 1 * * *"
   successfulJobsHistoryLimit: 1
   failedJobsHistoryLimit: 3
   concurrencyPolicy: Replace
   jobTemplate:
+    metadata:
+      labels:
+        app.kubernetes.io/name: svc-questionnaires
+        app.kubernetes.io/component: job
```
maxverbeek, line 17: For all of these, I don't know if it's a problem that every job shares the same matchLabels section. It could be thatthe controller will think pods from other jobs will belong to a different job. Should probably be fine though.

maxverbeek:
Keep in mind that matchLabels are immutable, and changing them will require deleting the old resource first. I'm not sure if ArgoCD can do this automatically

PR #3775, opened 2025-06-30
Making 'Satisfaction' question in the training log questionnaire optional

idiloks:
# Description of feature

"Satisfaction" question in the training log is now optional. When type = 'range' values do not have a value the name attribute is removed. Therefore, the browser does not assign a default value to the range questions.

### DIFF
--- a/projects/sport-data-valley/seeds/questionnaires/training_log.rb
+++ b/projects/sport-data-valley/seeds/questionnaires/training_log.rb
@@ -139,12 +139,12 @@ dagboek_content = [
   },
   {
     id: :v7,
-    title: { nl: 'Tevredenheid over training', en: 'Training satisfaction' },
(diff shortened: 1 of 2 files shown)


projects/sport-data-valley/seeds/questionnaires/training_log.rb
```diff
@@ -166,7 +167,7 @@
     title: { nl: 'Plezier tijdens training (optioneel)', en: 'Fun during training (optional)' },
     type: :range,
     min: 1,
-    max: 5,
+    max: 5,
```
emerencia, line 170: remove?

emerencia:
Specs are failing, can you look into which specs are going wrong and fix them?

emerencia:
The rest LGTM.

PR #3776, opened 2025-07-02
feat(coach-dashboard): added testing for the 'fun' element

idiloks:
# Description of feature
Updated the seeds to include values for 'fun' column therefore testing can be done via example data.


emerencia (approved):
LGTM

PR #3911, opened 2025-12-16
[AUKATI TIME][SVC QUESTIONNAIRES] Start hour dropdown from 12pm instead of 12am (survey settings)

nuisagumen:
Changed the am/pm logic as requested by Beth: starts at pm instead of am, removed 0 am as it does not exist. + display of sleeping hours

### DIFF
--- a/app/assets/javascripts/questionnaire_logic.js.erb
+++ b/app/assets/javascripts/questionnaire_logic.js.erb
@@ -142,6 +142,51 @@ function time_element() {
   }
 }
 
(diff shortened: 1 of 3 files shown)


pmkihara (approved):
LGTM!

I think we should wait for USC reply about the time dropdown before we merge this. We might want to add another parameter to check whether to display the time in the normal order or with pm before am if they want to keep the normal order for the wake time question.

emerencia (approved):
LGTM

PR #3978, opened 2026-03-03
feat:hide protocols for SDV SLA

idiloks:


lib/tasks/maintenance.rake
```diff
@@ -37,4 +39,29 @@ namespace :maintenance do
     end
     puts 'Scrambling people - done'
   end
+
+  desc 'Cancel active subscriptions for target protocols'
```
WSeubring: Should this be in this rake, or a thing on its own

WSeubring (approved):

PR #3993, opened 2026-03-16
Deploy Production 

idiloks:


app/controllers/api/v1/jwt_api/protocol_controller.rb
```diff
@@ -6,10 +6,18 @@ module JwtApi
       class ProtocolController < JwtApiController
         include ProtocolHelper
 
+        BLOCKED_PROTOCOL_NAMES = %w[
+          move_mood_motivation
```
maxverbeek, line 10: Not sure about this

maxverbeek (approved):
not sure about this in u can act

app/controllers/api/v1/jwt_api/protocol_controller.rb
idiloks: had to do this to hide the protocol names from the list, but still had to keep the data of the users who filled it beforehand. let me know if you have any suggestions. i cannot hide it in the front-end, because of the seeds it still shows up in the protocol list with its key names.

PR #3994, opened 2026-03-17
fix: upgrade Node.js from 18 to 20 in Dockerfile

emerencia:
The Dockerfile installs Node.js 18 via nodesource, but `package.json` requires `>= 20`. This causes `yarn install --check-files` to fail:

```
error u-can-act@0.1.0: The engine "node" is incompatible with this module. Expected version ">= 20". Got "18.20.8"
```

**Fix:** Change `setup_18.x` → `setup_20.x` in the Dockerfile.


idiloks:
lgtm!

PR #4007, opened 2026-03-27
fix(questionnaire): add explicit value keys to KCCQ and PROMIS likert options

WSeubring:
## Summary
- KCCQ and PROMIS likert options were defined as simple translation hashes (`{ nl: '...', en: '...' }`) without explicit `value` keys
- The likert generator falls back to using the Dutch display text as the HTML form value, causing responses to store text like `"Uitstekend"` instead of numeric `"1"`
- Added explicit `value` keys to all 33 affected likert options, matching the answer mappings from `kccq_answers.csv` and `promis_answers.csv`

## Changes
- **KCCQ**: 6 likert questions (v2–v7) — values `1`–`5` or `1`–`7`
- **PROMIS**: 27 likert questions (all sections) — values `1`–`5`
- Range-type questions unchanged (already submit numeric values)

## Impact
- No code changes to generators — they already support the `{ title: {...}, value: '...' }` format via `option[:raw][:value].presence || option[:raw][:title]`
- Other questionnaires (daily, RestQ, climbing_accident) are unaffected
- **Note**: Existing stored responses with Dutch text values will not be retroactively changed. After re-seeding, new responses will store numeric keys.

## Test plan
- [ ] Re-seed questionnaires on dev environment
- [ ] Fill in KCCQ and PROMIS questionnaires
- [ ] Verify responses contain numeric values (`"1"`, `"2"`, etc.) instead of Dutch text
- [ ] Verify exported data matches `kccq_answers.csv` / `promis_answers.csv` mappings
- [ ] Verify other questionnaires still work correctly

### DIFF
--- a/projects/sport-data-valley/seeds/questionnaires/kccq.rb
+++ b/projects/sport-data-valley/seeds/questionnaires/kccq.rb
@@ -86,11 +86,11 @@ dagboek_content = [
              en: '<br>2. Over the past 2 weeks, how many times did you have swelling in your feet, ankles or legs when you woke up in the morning?' },
     type: :likert,
     options: [
-      { nl: 'Elke morgen', en: 'Every morning' },
-      { nl: '3 keer of meer per week, maar niet elke dag', en: '3 or more times per week but not every day' },
-      { nl: '1 tot 2 keer per week', en: '1-2 times per week' },
-      { nl: 'Minder dan 1 keer per week', en: 'Less than once a week' },
-      { nl: 'Nooit gedurende de afgelopen 2 weken', en: 'Never over the past 2 weeks' }
+      { title: { nl: 'Elke morgen', en: 'Every morning' }, value: '1' },
+      { title: { nl: '3 keer of meer per week, maar niet elke dag', en: '3 or more times per week but not every day' }, value: '2' },
+      { title: { nl: '1 tot 2 keer per week', en: '1-2 times per week' }, value: '3' },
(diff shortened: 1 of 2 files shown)


projects/sport-data-valley/seeds/questionnaires/kccq.rb
```diff
@@ -86,11 +86,11 @@
              en: '<br>2. Over the past 2 weeks, how many times did you have swelling in your feet, ankles or legs when you woke up in the morning?' },
     type: :likert,
     options: [
-      { nl: 'Elke morgen', en: 'Every morning' },
-      { nl: '3 keer of meer per week, maar niet elke dag', en: '3 or more times per week but not every day' },
-      { nl: '1 tot 2 keer per week', en: '1-2 times per week' },
-      { nl: 'Minder dan 1 keer per week', en: 'Less than once a week' },
-      { nl: 'Nooit gedurende de afgelopen 2 weken', en: 'Never over the past 2 weeks' }
+      { title: { nl: 'Elke morgen', en: 'Every morning' }, value: '1' },
```
idiloks, line 89: the only thing i am worried about is that, we need to make sure amsterdam umc knows what answer corresponds to what value. because in this example it might steer them from the actual answer. for example the value 3 can be seen as 3 times a week, when "3 times a week" is actually 2. there are other questions like this. so that is why i dont know if this is the best way to go.

idiloks: wesley sent a document to amsterdam umc about which questionnaire corresponds to which value, so this is resolved

PR #4021, opened 2026-04-21
Feature/mr upgrade rails

m-rios:
# Description of feature
...

# Checklist


m-rios:
@emerencia Rails 7.0 and 7.1 made updated the algorithm used for computing digests from SHA1 to SHA256. This would affect signed cookies. Basically, any cookie created before the migration is invalidated because we now compute the digest differently. I am not sure how important cookies are here, afaik they are short lived session cookies. If that's the case, then we don't need to worry about this, but I wanted to check with you first. You can read more about the SHA conundrum here https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#key-generator-digest-class-change-requires-a-cookie-rotator

emerencia:
> @emerencia Rails 7.0 and 7.1 made updated the algorithm used for computing digests from SHA1 to SHA256. This would affect signed cookies. Basically, any cookie created before the migration is invalidated because we now compute the digest differently. I am not sure how important cookies are here, afaik they are short lived session cookies. If that's the case, then we don't need to worry about this, but I wanted to check with you first. You can read more about the SHA conundrum here https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#key-generator-digest-class-change-requires-a-cookie-rotator

Yes they are short-lived cookies. It's fine if they expire. I think they are created on-the-fly when logging in with a JWT token or with the token that's in the questionnaire invitation URL.

lib/docker-entrypoint.sh
```diff
@@ -9,6 +9,12 @@ if [ -f /app/tmp/pids/server.pid ]; then
 fi
 
 cd /app
+
+# Unfreeze Gemfile.lock if running on development mode so we can install new gems
+if [ "$RESET_DB_ON_FAIL" == "true" ]; then
+  bundle config unset frozen
+fi
+
```
emerencia, line 17: nice

Gemfile
```diff
@@ -210,4 +209,5 @@ gem 'warden-jwt_auth'
 
 # Reduces boot times through caching; required in config/boot.rb
 gem 'bootsnap', require: false
+gem 'cgi', '< 0.5'
```
emerencia: Can you add a comment why this is needed?

emerencia (approved):
LGTM. Specs are failing though.

app/controllers/api/v1/basic_auth_api/protocols_controller.rb
```diff
@@ -14,17 +14,18 @@ def create
...
+                        :duration,
+                        :invitation_text,
+                        :informed_consent_questionnaire_key,
+                        { questionnaires:
+                          [:key,
+                           { measurement:
+                             %i[open_from_offset open_from_day period open_duration reminder_delay priority
+                                stop_measurement should_invite only_redirect_if_nothing_else_ready prefilled
+                                redirect_url collapse_duplicates] }] },
+                        { push_subscriptions: %i[name url method] }]]
```
emerencia: 🐛 Likely contract change. The original `params.require(:protocol).permit([...])` expected `params[:protocol]` to be a single hash. The new `params.expect(protocol: [[ ... ]])` (note the **double** brackets) declares `protocol` as an *array of hashes* and returns an array, not a single `ActionController::Parameters` object.

Downstream `CreateOrUpdateProtocol.run(**update_protocol_params)` will then double-splat an array, which will blow up at runtime with `TypeError: no implicit conversion of Array into Hash` (or similar).

If the intent was a single object, drop one set of brackets:

```ruby
params.expect(
  protocol: [:name,
             :duration,
             :invitation_text,
             :informed_consent_questionnaire_key,
             { questionnaires: [:key, { measurement: %i[...] }] },
             { push_subscriptions: %i[name url method] }]
)
```

Also worth adding a request spec that actually hits this endpoint — the rest of the controller-level specs in this PR didn't catch it.

config/initializers/knock.rb
```diff
@@ -15,51 +19,47 @@
   ## Configure the audience claim to identify the recipients that the token
   ## is intended for.
   ## If using Auth0:
-  if ENV['AUTH0_CLIENT_ID'].present? && Rails.application.secrets.auth0_client_id.present?
-    config.token_audience = -> { Rails.application.secrets.auth0_client_id }
+  if ENV['AUTH0_CLIENT_ID'].present? && ENV['AUTH0_CLIENT_ID'].present?
```
emerencia: Duplicate condition: `ENV['AUTH0_CLIENT_ID'].present? && ENV['AUTH0_CLIENT_ID'].present?`. Looks like a search-and-replace that overshot — the second clause was previously `Rails.application.secrets.auth0_client_id.present?`. Since `Rails.application.secrets` is gone, this whole line can collapse to:

```ruby
if ENV['AUTH0_CLIENT_ID'].present?
```

config/initializers/knock.rb
```diff
@@ -15,51 +19,47 @@
...
       OpenSSL::PKey::RSA.new(
-        OpenSSL::X509::Certificate.new(Base64.strict_decode64(Rails.application.secrets.signing_certificate))
+        OpenSSL::X509::Certificate.new(Base64.strict_decode64(ENV['AUTH0_SIGNING_CERTIFICATE']))
       )
     }
   else
-    config.token_secret_signature_key = -> {
-      Rails.application.secrets.signing_certificate
+    config.token_secret_signature_key = lambda {
+      ENV['AUTH0_SIGNING_CERTIFICATE']
```
emerencia, line 45: Nit: ENV access style is inconsistent in this file — `ENV.fetch('TOKEN_SIGNATURE_ALGORITHM', nil)` on line 31 vs. `ENV['AUTH0_SIGNING_CERTIFICATE']` here and on lines 40/59/63. Pick one. (Rubocop's `Style/FetchEnvVar` cop will yell about the bare `ENV[]` form.)

app/controllers/languages_controller.rb
```diff
@@ -23,7 +23,7 @@ def change
     end
 
     flash.notice = I18n.t('pages.languages.flash_messages.notice.locale_updated')
-    return redirect_to change_params[:cb] if params[:cb].present?
+    return redirect_to change_params[:cb], allow_other_host: true if params[:cb].present?
```
emerencia, line 26: ⚠️ `change_params[:cb]` is a fully user-controlled query parameter and is now redirected to with `allow_other_host: true`. That's an open redirect — an attacker can craft `…/language/change?locale=nl&cb=https://evil.example.com` and we'll happily forward the user's session there.

If cross-host redirects are a real requirement (e.g. callback URLs into other u-can-act apps), please add an allowlist check before the redirect, e.g.:

```ruby
allowed = URI.parse(change_params[:cb]).host.in?(Rails.application.config.settings.allowed_redirect_hosts)
return redirect_to(change_params[:cb], allow_other_host: true) if allowed
```

Same concern applies to `people_controller` and `questionnaire_controller`, which now also pass `allow_other_host: true` for `NextPageFinder.get_next_page` results that ultimately come from `measurement.redirect_url` / `callback_url` params (the new `redirects to an external callback_url when one is supplied` spec confirms `https://example.com/completed` is reachable).

app/models/auth_user.rb
```diff
@@ -79,10 +79,10 @@ def access_level_from_payload(payload)
       access_level_admin = metadata_from_payload(payload)['access_level']&.include?(ADMIN_ACCESS_LEVEL)
 
       if role_admin
-        ActiveSupport::Deprecation
-          .warn('roles should not be used in the payload anymore. ' \
-                'This was renamed to access_level. ' \
-                'Please update the payload.')
+        ActiveSupport::Deprecation.new
```
emerencia, line 82: `ActiveSupport::Deprecation.new.warn(...)` instantiates a brand-new `Deprecation` instance with no `gem_name` / `deprecation_horizon`, and disconnected from `Rails.application.deprecators`. The output won't be silenceable in tests via the usual `ActiveSupport::Deprecation.behavior=` either.

In Rails 7.1+ the recommended pattern is to register a deprecator once and reuse it:

```ruby
# config/initializers/deprecators.rb
Rails.application.deprecators[:u_can_act] = ActiveSupport::Deprecation.new('next', 'u-can-act')

# usage
Rails.application.deprecators[:u_can_act].warn('roles should not be used …')
```

The corresponding spec change to `expect_any_instance_of(ActiveSupport::Deprecation)` is a bit of a smell — it works only because every call gets a fresh instance.

app/models/questionnaire.rb
```diff
@@ -18,7 +18,8 @@ class Questionnaire < ApplicationRecord
   validates :name, presence: true, uniqueness: true
   validates :content, presence: true
   validates :key, presence: true, uniqueness: true, format: { with: /\A[a-z_0-9]+\Z/ }
-  serialize :content # Don't specify Hash type because otherwise databases with existing questionnaires won't work
+  # Don't specify Hash type because otherwise databases with existing questionnaires won't work
+  serialize :content, coder: YAML
```
emerencia, line 22: Switching to `serialize :content, coder: YAML` is the right Rails 7.1+ move, but: in Rails 7.1+ `coder: YAML` defaults to `YAML.safe_load` with a restricted permitted-classes list (basically `Symbol`, `Date`, `Time`, `BigDecimal`, etc. are *not* allowed unless you opt in via `safe_load: true, permitted_classes: [...]`).

Given the comment says existing DBs already have serialized questionnaires, please double-check that loading an existing prod-shaped row still works. If it doesn't, you'll want either:

```ruby
serialize :content, coder: YAML, type: nil # explicit; same default
# or
serialize :content, coder: YAML.unsafe_method_or_custom_coder
```

or a custom coder that calls `YAML.unsafe_load`. A spec that round-trips a representative production-shaped questionnaire would be reassuring.

Gemfile
```diff
@@ -210,4 +209,5 @@ gem 'warden-jwt_auth'
...
 gem 'logger', '~> 1.6.0' # logger 1.7+ breaks Rails 6.1
```
emerencia: Comment is now stale — we just upgraded *past* Rails 6.1, all the way to 8.1. Either drop the pin (let Bundler resolve `~> 1.7` / latest with Rails 8) or update the comment to whatever the current incompatibility actually is.

emerencia (commented):
Reviewed the Rails 6.1 → 8.1 upgrade. Big jump and most of the mechanical changes look right (strong params `expect`, `allow_other_host`, `to_fs`, deprecator API, Active Storage migrations, `serialize ..., coder: YAML`). A few items that look like real issues, plus a handful of clean-up nits — see inline.

High-priority:
- `protocols_controller#update_protocol_params` now uses `expect(protocol: [[ ... ]])` (array of hashes) where the original `require(:protocol).permit(...)` expected a single hash. Looks like an unintended contract change.
- `application_controller`: `before_action :work_off_delayed_jobs` was removed and `config/initializers/workless.rb` is gone, but I don't see a replacement strategy in this PR. How are delayed jobs going to run after this merge (worker process, cron, something else)?
- `languages_controller#change` now passes `change_params[:cb]` (user-controlled) to `redirect_to ..., allow_other_host: true`. That's an open-redirect; one of the new specs in `questionnaire_controller_spec.rb` even codifies redirecting to `https://example.com/completed`. Worth confirming this is intentional and, if so, allowlisting the hosts.
- `knock.rb`: `if ENV['AUTH0_CLIENT_ID'].present? && ENV['AUTH0_CLIENT_ID'].present?` — duplicate condition (the second clause used to be `Rails.application.secrets.auth0_client_id`).

Lower priority cleanup:
- Stale Gemfile comments (`logger 1.7+ breaks Rails 6.1`, `not compatible with rails 7.2`).
- `gem 'cgi', '< 0.5'` and the new `config/initializers/json_schema.rb` could use a one-line comment explaining why.
- `auth_user.rb` calls `ActiveSupport::Deprecation.new.warn(...)` — that creates a fresh, unconfigured deprecator. Better to either keep using a class-level deprecator or register one via `Rails.application.deprecators[:u_can_act]`.
- `api_controller` adds `authenticate_auth_user` / `current_auth_user` but I don't see them being called anywhere — leftover from an earlier draft?
- `bin/questionnaire_dashboard_seeds.rb` has only a file-mode change (0644 → 0755) — looks accidental.
- PR is currently `mergeable_state: dirty`, will need a rebase against `develop` before merge. Don't forget to run the new Active Storage migrations on staging/production.

Once the protocols_controller bug, the workless replacement, and the open-redirect surface are addressed, the rest looks reasonable for an 8.1 jump. — review by Computer (Claude/Opus) on Ando's behalf.

Gemfile
```diff
@@ -28,7 +28,8 @@ gem 'haml-rails'
 gem 'uglifier'
 
 # Use React for the UI
-gem 'react-rails'
+# Earlier versions are no longer compatible with rails 7.2 because of connection_pool API change
```
emerencia, line 31: Nit: comment says `rails 7.2` but this PR jumps to Rails 8.1. Either update to mention 8.x or just simplify to `# Pinned: needs the connection_pool API introduced in 3.3.0`.

Gemfile
emerencia: Could you add a one-line comment for `gem 'cgi', '< 0.5'`? Future-us will wonder why we're capping it (presumably the API change in cgi 0.5 broke something — Rack/Rails interop?). A link to the relevant issue/PR is enough.

app/controllers/api/v1/api_controller.rb
```diff
@@ -16,6 +16,14 @@ def unauthorized_entity(entity_name)
                status: :unauthorized
       end
 
+      def authenticate_auth_user
+        unauthorized_entity('auth_user') unless authenticate_entity('auth_user')
+      end
+
+      def current_auth_user
+        authenticate_entity('auth_user')
+      end
```
emerencia, line 27: These two helpers (`authenticate_auth_user` / `current_auth_user`) are added but I don't see anything in this PR that calls them. If they're prep for an upcoming change, fine; otherwise this is dead code worth removing.

config/initializers/json_schema.rb
```diff
@@ -0,0 +1,3 @@
+# frozen_string_literal: true
+
+JSON::Validator.use_multi_json = false
```
emerencia, line 4: Mind adding a one-line comment for context? Something like `# Disable multi_json now that JSON::Validator deprecates it (see <link>)` so the next person doesn't have to git-blame to understand the line.

app/models/auth_user.rb
m-rios: choosing to ignore this, we don't have a horizon, and we don't really want this to be silenceable. NOt worth the effort

Gemfile
m-rios: Ignoring this, I want to know which rails version made us introduce this

PR #4026, opened 2026-05-04
fix: production build error due to unloaded gem

m-rios:
# Description of feature
...

# Checklist

### DIFF
--- a/config/initializers/json_schema.rb
+++ b/config/initializers/json_schema.rb
@@ -1,4 +1,12 @@
 # frozen_string_literal: true
 
+# json-schema is a transient dependency coming from rswag. Rswag is only
+# available on the development and test environments. If we run this initializer
+# on production we get a boot fail. We could just skip it on production, but
+# actually checking for the presence of the gem is more correct.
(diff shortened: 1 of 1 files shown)


millenevprado (approved):

config/initializers/json_schema.rb
```diff
@@ -1,4 +1,12 @@
 # frozen_string_literal: true
 
+# json-schema is a transient dependency coming from rswag. Rswag is only
+# available on the development and test environments. If we run this initializer
+# on production we get a boot fail. We could just skip it on production, but
+# actually checking for the presence of the gem is more correct.
```
pmkihara, line 6: I like this approach. Maybe just a small comment: json-schema is also a dependency from mcp, which is a dependency of rubocop (only available in the test environment), so checking if the gem has loaded is more correct indeed.

pmkihara:
The rest lgtm, but Millene had already reviewed it while I was checking for dependencies

config/initializers/json_schema.rb
m-rios: :raised_hands:

